PR #3, opened 2015-04-08
Add Support for Number Insight and Verify

waleo:


daniele-athome:
@tjlytle I think you can close this right?

PR #6, opened 2015-08-31
Maven Support

jbcpollak:
This is a new PR for Maven support - note it is configured for Java 1.7 - you can easily change this in the pom.xml, just find the `<source>` and `<target>` tags and change the value from 1.7 to 1.6 - the code as is uses the `<>` operator, so Java 1.6 isn't supported.


jbcpollak:
Two more notes:

* I wasn't able to test the unit tests because I didn't put my api key info into the test.properties file, but it should work fine. - You will need to skip the tests in order to build the jar if you can't provide a test.properties file.
* The source code has been relocated to `src/main/java`, which is the Maven convention. I did update the build.xml also, so it should still build with Ant.

judy2k:
Hello! Thanks for your contribution to the Nexmo Java client library! Sorry we didn't get to this sooner.

We've been doing lots of work in the past 2 months on a branch called `version-2`, which adds support for Maven POM generation from Gradle and 1.6 compatibility, among other things. So, I'm going to close this PR.

Maintenance of this project is going to be more active from now on, so hopefully future PRs will be managed more quickly than yours has been.

Issue #10, opened 2016-04-22
Add to Maven


leggetter:

judy2k:
This has been resolved by the first `2.0.0-SNAPSHOT` release.

PR #14, opened 2016-10-16
Gradlify build + add Maven upload task

### LINKED ISSUE #11: Build with Gradle

andyberry88:
- Adds a Gradle build file to build, test and upload to Maven. Fixes #11 and goes part way to fixing #10.
- Uses the Gradle wrapper so Gradle doesn't need to be installed locally - use `gradlew build`
- Still needs some tidy up which I haven't had time to finish
  - Readme needs updating
  - Some tests are failing and have been `@Ignore`d - I think because they expect an API user/key so they need refactoring to do something like mock out the HTTP requests/responses
  - The Maven upload needs to be tested + POM info checked


judy2k:
I'm just looking at this and checking the tests now.

judy2k:
Replaced by #15

PR #16, opened 2016-10-17
Mocking in tests

judy2k:
Stubs out all tests to the API so that tests can be run without any network calls. These are not unit tests per se, they are mostly full-stack calls but without any actual HTTP calls being made.

Some refactoring is also being conducted, in order to make the code more testable, and I also swapped out some throwing of `Exception` for more specific exception subclasses.

**Backwards Incompatibility**:
- `NexmoSmsClient.submitMessage` no longer supports networkCode and reachabilityCheck.
- Failed requests no longer automatically retry for insights, verification, or messaging.


judy2k:
These changes have been incorporated into the `version-2` branch, so I'm going to close this PR.

PR #18, opened 2016-12-08
Add 2 VAPI endpoints

ChrisGuzman:
- [x] Add a call modifier endpoint
- [x] Add the DTMF endpoint

### DIFF
--- a/src/main/java/com/nexmo/client/dtmf/DTMFEndpoint.java
+++ b/src/main/java/com/nexmo/client/dtmf/DTMFEndpoint.java
@@ -1,4 +1,4 @@
-package com.nexmo.client.voice;/*
+package com.nexmo.client.dtmf;/*
  * Copyright (c) 2011-2016 Nexmo Inc
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -20,14 +20,22 @@ package com.nexmo.client.voice;/*
  * THE SOFTWARE.
  */
 
-public class Payload {
-    private String action;
+import com.nexmo.client.HttpWrapper;
+import com.nexmo.client.NexmoClientException;
+import com.nexmo.client.voice.DTMFRequest;
+import com.nexmo.client.voice.DTMFResponse;
+import com.nexmo.client.voice.SendDTMFMethod;
 
-    public Payload(String action) {
-        this.action = action;
+import java.io.IOException;
+
+public class DTMFEndpoint {
+    private final SendDTMFMethod sendDTMF;
+
+    public DTMFEndpoint(HttpWrapper httpWrapper) {
+        this.sendDTMF = new SendDTMFMethod(httpWrapper);
     }
 
-    public String getAction() {
-        return action;
+    public DTMFResponse put(DTMFRequest dtmfRequest) throws IOException, NexmoClientException {
+        return this.sendDTMF.execute(dtmfRequest);
     }
 }
--- a/src/main/java/com/nexmo/client/voice/CallModifier.java
+++ b/src/main/java/com/nexmo/client/voice/CallModifier.java
@@ -27,25 +27,29 @@ import com.nexmo.client.NexmoUnexpectedException;
  */
 public class CallModifier {
     private String uuid;
-    private Payload payload;
+    private ModifyCallPayload modifyCallPayload;
 
-    public CallModifier(String uuid, Payload payload) {
+    public CallModifier(String uuid, String action) {
         this.uuid = uuid;
-        this.payload = payload;
+        this.modifyCallPayload = new ModifyCallPayload(action);
     }
 
     public String getUuid() {
         return uuid;
     }
 
-    public Payload getPayload() {
-        return payload;
+    public String getAction() {
+        return modifyCallPayload.getAction();
+    }
+
+    public void setAction(String action) {
+        this.modifyCallPayload.setAction(action);
     }
 
     public String toJson() {
         try {
             ObjectMapper mapper = new ObjectMapper();
-            return mapper.writeValueAsString(this.payload);
+            return mapper.writeValueAsString(this.modifyCallPayload);
         } catch (JsonProcessingException jpe) {
             throw new NexmoUnexpectedException("Failed to produce json from CallModifier object.", jpe);
         }
--- a/src/main/java/com/nexmo/client/voice/CallModifier.java
+++ b/src/main/java/com/nexmo/client/voice/CallModifier.java
@@ -1,0 +1,53 @@
+package com.nexmo.client.voice;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.nexmo.client.NexmoUnexpectedException;
+
+/*
+ * Copyright (c) 2011-2016 Nexmo Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+public class CallModifier {
+    private String uuid;
+    private Payload payload;
+
+    public CallModifier(String uuid, Payload payload) {
+        this.uuid = uuid;
+        this.payload = payload;
+    }
+
+    public String getUuid() {
+        return uuid;
+    }
+
+    public Payload getPayload() {
+        return payload;
+    }
+
+    public String toJson() {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            return mapper.writeValueAsString(this.payload);
+        } catch (JsonProcessingException jpe) {
+            throw new NexmoUnexpectedException("Failed to produce json from Call object.", jpe);
+        }
+    }
+}
--- a/src/main/java/com/nexmo/client/voice/CallRecord.java
+++ b/src/main/java/com/nexmo/client/voice/CallRecord.java
@@ -178,7 +178,7 @@ public class CallRecord {
             mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
             return mapper.readValue(json, CallRecord.class);
         } catch (IOException jpe) {
-            throw new NexmoUnexpectedException("Failed to produce json from Call object.", jpe);
+            throw new NexmoUnexpectedException("Failed to produce json from CallRecord object.", jpe);
         }
     }
 }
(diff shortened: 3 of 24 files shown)


src/main/java/com/nexmo/client/dtmf/SendDTMFMethod.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.util.EntityUtils;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+
+public class SendDTMFMethod extends AbstractMethod<DTMFRequest, DTMFResponse> {
```
ChrisGuzman: @judy2k I copied your pattern of extending `AbstractMethod` from the voice client methods. Is there another way I should do this?

src/main/java/com/nexmo/client/dtmf/DTMFPayload.java
```diff
@@ -0,0 +1,37 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+public class DTMFPayload {
+    private String digits;
```
ChrisGuzman: The [DTMF docs](https://docs.nexmo.com/voice/voice-api/api-reference#dtmfparameters) says that it takes "The array of digits to send to the Call" but looking at the implementations for ruby and python, it looks like they're sending a string. Do you know which is correct?

src/main/java/com/nexmo/client/dtmf/SendDTMFMethod.java
judy2k: I think that's correct in this case. You'd only want to implement your own Method from scratch if you wanted to bypass the authentication mechanism, or do something else weird ;-)

src/main/java/com/nexmo/client/dtmf/DTMFPayload.java
judy2k: I'd follow the existing implementations, and maybe file a documentation ticket. The implementations have the advantage of being tested :wink:

ChrisGuzman:
@judy2k I've noticed that in the `Call.java` class the `to` and `answerUrl` are Strings not Lists are the API page suggests https://docs.nexmo.com/voice/voice-api/api-reference#ccparameters. I haven't tested this branch against the API but wanted to bring it to your attention. Maybe we need to change the attributes in the `Call.java` class?

leggetter:
Is the workflow and signatures of other APIs/resources/sub-resources similar to this? e.g.

1. Create a request object
2. Upon creation pass a payload object with the optional request parameters?
3. Once you've made the request you get a response object

leggetter:
Would it be possible to update the README to show what the VAPI functionality is going to look like? I appreciate that may change, but I'd just like to get an idea of what we're likely aiming for.

ChrisGuzman:
@leggetter asked:
> Is the workflow and signatures of other APIs/resources/sub-resources similar to this? e.g.
> Create a request object
> Upon creation pass a payload object with the optional request parameters?
> Once you've made the request you get a response object

In my experience, yes most SDKs for APIs have a similar workflow.

judy2k:
It's a limitation of the language. Java doesn't have default arguments for params, or dictionary literals, so depending on the complexity of the call, if you have any optional arguments you either provide multiple overloaded methods, or you provide a configurable request object.

The whole thing can be made slightly nicer by adding Builders for the request objects, but I haven't done that (yet - I'm probably going to).

ChrisGuzman:
@judy2k I'm at -2.71% code coverage. Can you help me with the last bit of tests I should write or should we merge it and worry about adding additional tests later.

ChrisGuzman:
Actually before we merge it I guess it would be good to do a proper code review. I think some of the models (eg: `Payload.java`) and some of the methods in the tests could be better named.

src/main/java/com/nexmo/client/dtmf/DTMFEndpoint.java
```diff
@@ -0,0 +1,41 @@
+package com.nexmo.client.dtmf;/*
```
judy2k: I think everything related to dtmf should be under the voice package. I don't think dtmf code makes sense outside the context of VAPI? (Specifically, I think this should be `com.nexmo.client.voice.dtmf`)

src/main/java/com/nexmo/client/voice/DTMFPayload.java
```diff
@@ -0,0 +1,37 @@
+package com.nexmo.client.voice;/*
```
judy2k: The formatting of these copyright comments starting on the same line as the package declaration is a mistake in my IntelliJ template. Feel free to put them before the package declaration if you prefer - I think this is what we agreed the other day.

src/main/java/com/nexmo/client/voice/DTMFRequest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.nexmo.client.NexmoUnexpectedException;
+
+public class DTMFRequest {
+    private String uuid;
+    private DTMFPayload payload;
+
+    public DTMFRequest(String uuid, DTMFPayload payload) {
```
judy2k: Would it be worth adding a convenient constructor with the following signature?

```java
public DTMFRequest(String uuid, String digits)
```

I figure that would be nicer for the user, given that the DTMFPayload only really exists as an implementation detail, for JSON serialization - in fact, maybe remove this constructor entirely and have the DTMFPayload object initialized inside the new constructor?

src/main/java/com/nexmo/client/voice/NexmoVoiceClient.java
```diff
@@ -48,4 +51,13 @@ public CallRecordPage listCalls() throws IOException, NexmoClientException {
     public CallRecordPage listCalls(CallsFilter filter) throws IOException, NexmoClientException {
         return calls.get(filter);
     }
+
+    public void sendDTMF(DTMFRequest dtmfRequest) throws IOException, NexmoClientException {
```
judy2k: I think maybe either replace or overload this method with the signature below? What do you think?

```java
sendDTMF(String uuid, String digits)
```

src/main/java/com/nexmo/client/voice/Payload.java
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+public class Payload {
```
judy2k: The name `Payload` is very generic - every endpoint that accepts `PUT` or `POST` requests takes a payload. Maybe `ModifyCallPayload`?

src/main/java/com/nexmo/client/voice/CallModifier.java
```diff
@@ -0,0 +1,53 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+public class CallModifier {
+    private String uuid;
+    private Payload payload;
+
+    public CallModifier(String uuid, Payload payload) {
```
judy2k: I feel the same way here about `Payload` as I do about `DTMFPayload` - it's an internal detail, and ideally we should hide it from the user. `CallModifier(String uuid, String action)` instead?

I was wondering if maybe we should have a package for internal-only objects. What do you think is the best way to "hide" them from the user?

src/main/java/com/nexmo/client/dtmf/DTMFEndpoint.java
ChrisGuzman: Agreed

src/main/java/com/nexmo/client/voice/DTMFPayload.java
ChrisGuzman: Will change

src/main/java/com/nexmo/client/voice/Payload.java
ChrisGuzman: `ModifyCallPayload` sounds good

src/main/java/com/nexmo/client/voice/DTMFRequest.java
ChrisGuzman: Yes I think you're right

ChrisGuzman:
@judy2k Can you review changes again and approve if everything's fine?

src/main/java/com/nexmo/client/voice/CallRecord.java
```diff
@@ -178,7 +178,7 @@ public static CallRecord fromJson(String json) {
             mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
             return mapper.readValue(json, CallRecord.class);
         } catch (IOException jpe) {
-            throw new NexmoUnexpectedException("Failed to produce json from Call object.", jpe);
+            throw new NexmoUnexpectedException("Failed to produce json from CallRecord object.", jpe);
```
judy2k: Nice catch.

src/main/java/com/nexmo/client/voice/SendDTMFMethod.java
```diff
@@ -0,0 +1,74 @@
...
+        put.setHeader("Content-Type", "application/json");
+        put.setEntity(new StringEntity(request.toJson()));
+        LOG.info("Request: " + request.toJson());
+        return put;
+    }
+
+    @Override
+    public DTMFResponse parseResponse(HttpResponse response) throws IOException {
+        String json = EntityUtils.toString(response.getEntity());
+        LOG.info("Received: " + json);
```
judy2k: I've been changing my log messages like this to DEBUG, but it's a task of its own, so feel free to leave this now and we can tidy up in a later PR.

src/test/java/com/nexmo/client/voice/CallModifierTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    public void setAction() throws Exception {
+        //this is not a real action
+        callModifier.setAction("pause");
+        assertEquals("pause", callModifier.getAction());
+    }
+
+    @Test
+    public void toJson() throws Exception {
+        String jsonString = "{" + "\"action\":\"hangup\"" + "}";
```
judy2k: What's the reason for defining this as three string literals?

src/test/java/com/nexmo/client/voice/endpoints/ModifyCallMethodTest.java
```diff
@@ -0,0 +1,101 @@
+package com.nexmo.client.voice.endpoints;
+
+import com.auth0.jwt.internal.com.fasterxml.jackson.databind.JsonNode;
```
judy2k: Be really careful about this - we have two different Jacksons available - one is in `com.fasterxml.jackson.databind.JsonNode`, and the other vendored into `com.auth0.jwt.internal.com.fasterxml.jackson.databind.JsonNode`. (My IDE kept auto-completing to the wrong one)

src/test/java/com/nexmo/client/voice/CallModifierTest.java
ChrisGuzman: I think it was a bad copy/paste. I'll clean it up

src/test/java/com/nexmo/client/voice/endpoints/ModifyCallMethodTest.java
ChrisGuzman: Doh

PR #19, opened 2016-12-08
1.6 compatibility

judy2k:
Made some minor changes to allow 1.6 compatibility (at least source compatibility - I haven't tested against the 1.6 classpath). This gives us compatibility with any JVM from the past 10 years, which should be more than enough, and the cost is not enormous.

If we decide to move to 1.7+ (which is the past 5 years), we can make that decision further down the line.

### DIFF
--- a/src/main/java/com/nexmo/messaging/sdk/callback/AbstractMOServlet.java
+++ b/src/main/java/com/nexmo/messaging/sdk/callback/AbstractMOServlet.java
@@ -165,10 +165,9 @@ public abstract class AbstractMOServlet extends HttpServlet {
             this.consumer.execute(task);
 
             // immediately ack the receipt
-            try (PrintWriter out = response.getWriter()) {
-                out.print("OK");
-                out.flush();
-            }
+            PrintWriter out = response.getWriter();
+            out.print("OK");
+            out.flush();
         } catch (NexmoCallbackRequestValidationException exc) {
             // TODO: Log this - it's mainly for our own use!
             response.sendError(400, exc.getMessage());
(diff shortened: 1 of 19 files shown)


judy2k:
I've taken the testing out of Travis, so we will get compiler errors if we use 1.7+ features, but no guarantees that the code will run under 1.6.

> JDK 7 is backwards compatible, we think it’s time for all projects to start testing against JDK 7 first and JDK 6 if resources permit. (from Travis docs)

src/main/java/com/nexmo/messaging/sdk/callback/AbstractMOServlet.java
```diff
@@ -165,10 +165,9 @@ private void handleRequest(HttpServletRequest request, HttpServletResponse respo
             this.consumer.execute(task);
 
             // immediately ack the receipt
-            try (PrintWriter out = response.getWriter()) {
```
ChrisGuzman: Why did you remove the `try` block here?

judy2k: The driver for the change was that the try (x) { blah; } syntax is 1.7+.

I should probably put in a try/catch. I wasn't aware the new syntax enforced a call to `close` in this case. Thanks for the prompt!

ChrisGuzman:
@judy2k Does this need another review?

judy2k:
@ChrisGuzman No, it needs a change, but it's low priority!

judy2k:
We don't have a dependency on this, and the PR is stale, so closing.

PR #20, opened 2017-01-13
Increase voice test coverage

judy2k:
Sorry - kept doing more little things in this PR until it ended up being quite big.

* Improving the test coverage of CallRecord and CallFilter.
* In doing so, realised that the new CallRecord object  (copied from `Call.java`) didn't really reflect the data coming back from the list-calls endpoint, so fixed it.
* I've also switched the attributes being inappropriately deserialized to `String`s to `Date` and enum types.
* Implemented "GET https://api.nexmo.com/v1/calls/{uuid}"


src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -28,10 +28,11 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nexmo.client.NexmoUnexpectedException;
 
-// TODO: Re-insert 'rate' and 'price' (currently being ignored)
-// TODO: Convert direction and status to enum values
+/**
+ * Call encapsulates the information required to create a call using {@link NexmoVoiceClient#createCall(Call)}
+ */
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
-@JsonIgnoreProperties({ "_links", "rate", "price", "duration", "start_time", "end_time" })
```
ChrisGuzman: Can a user still access the rate, price, duration, and start time?

src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -40,16 +41,9 @@
...
-    private String status = null;
-    private String direction = null;
+    private MachineDetection machineDetection = null;
     private Integer lengthTimer = null;
     private Integer ringingTimer = null;
-    private Integer duration = null;
-    private String network = null;
-
-    private String callId = null;
-    private String conversationId = null;
```
ChrisGuzman: Did we remove the ability to get the uuid or conversation id?

src/main/java/com/nexmo/client/voice/CallRecord.java
```diff
@@ -185,23 +161,24 @@ public void setNetwork(String network) {
         this.network = network;
     }
 
-    public String toJson() {
```
ChrisGuzman: Why remove this? Because a CallRecord is only returned from an API call and it's not something that we attach as a payload when making an API request?

src/main/java/com/nexmo/client/voice/Endpoint.java
```diff
@@ -29,7 +31,8 @@
     private String number;
     private String dtmfAnswer = null;
 
-    public Endpoint(){}
+    public Endpoint() {
+    }
```
ChrisGuzman: Is this a style preference?

src/main/java/com/nexmo/client/voice/NexmoVoiceClient.java
```diff
@@ -37,8 +37,8 @@ public NexmoVoiceClient(HttpWrapper httpWrapper) {
         calls = new CallsEndpoint(httpWrapper);
     }
 
-    public void createCall(Call callRequest) throws IOException, NexmoClientException {
```
ChrisGuzman: Should we expect that all methods in `NexmoVoiceClient` to return a type and not be `void`?

src/test/java/com/nexmo/client/voice/NexmoVoiceClientTest.java
```diff
@@ -0,0 +1,164 @@
...
+        when(client.execute(any(HttpUriRequest.class))).thenReturn(response);
+        when(entity.getContent()).thenReturn(new ByteArrayInputStream(content.getBytes("UTF-8")));
+        when(sl.getStatusCode()).thenReturn(statusCode);
+        when(response.getStatusLine()).thenReturn(sl);
+        when(response.getEntity()).thenReturn(entity);
+
+        byte[] keyBytes = testUtils.loadKey("test/keys/application_key");
+        AuthCollection authCollection = new AuthCollection();
+        authCollection.add(new JWTAuthMethod(
+                "951614e0-eec4-4087-a6b1-3f4c2f169cb0",
```
ChrisGuzman: Where did this JWT come from?

src/main/java/com/nexmo/client/voice/Call.java
judy2k: Yes - I've just told it to deserialize from JSON now, where's those values were always `null` before.

src/main/java/com/nexmo/client/voice/Call.java
judy2k: I renamed those values back to their JSON attribute names:

* `callId` ➡️ `uuid`
* `conversationId` ➡️ `conversationUuid`

I don't like these names because they're implementation-specific - they're just IDs, dammit! The fact that they're generated as UUIDs is irrelevant :)

src/main/java/com/nexmo/client/voice/CallRecord.java
judy2k: Yup.

src/main/java/com/nexmo/client/voice/Endpoint.java
judy2k: TBH it was probably reformatted automatically by IntelliJ, but yes. I figure we should follow the (Google Java)[https://google.github.io/styleguide/javaguide.html#s4.1-braces] guidelines unless you have a better idea? (It looks like the Sun/Oracle styleguide has all but disappeared while I've been away from Java)

Also, 4-spaces-per-tab, not 2!

judy2k: I should actually stick something in the README to this effect in a "Contribute" section.

src/main/java/com/nexmo/client/voice/NexmoVoiceClient.java
judy2k: Yes? This line's pink because it's been removed?

src/test/java/com/nexmo/client/voice/NexmoVoiceClientTest.java
judy2k: It's an application I created, but without the key no-one can abuse it, so the fact this ID is valid is (mostly) unimportant.

src/main/java/com/nexmo/client/voice/Endpoint.java
ChrisGuzman: Yeah that's a good idea. I'll update my style in intellij

PR #23, opened 2017-01-19
Ncco objects

judy2k:
Implementing NCCO objects before implementing server-side code to handle webhook requests.


judy2k:
Dammit, I missed that eventUrl is always a list of URLs.

src/main/java/com/nexmo/client/voice/ncco/InputNcco.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public String getAction() {
+        return ACTION;
+    }
+
+    @Override
+    public String toJson() {
+        return NccoSerializer.getInstance().serializeNcco(this);
```
ChrisGuzman: Smart. I like this

src/main/java/com/nexmo/client/voice/ncco/ConnectNcco.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public Integer getTimeout() {
+        return timeout;
+    }
+
+    public void setTimeout(Integer timeout) {
+        this.timeout = timeout;
+    }
+
+    public Integer getLimit() {
```
ChrisGuzman: Will the API sometimes not return this field?

src/main/java/com/nexmo/client/voice/ncco/ConnectNcco.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public String getFrom() {
+        return from;
+    }
+
+    public void setFrom(String from) {
+        this.from = from;
+    }
+
+    public Integer getTimeout() {
```
ChrisGuzman: Will the API sometimes not return this field?

PR #24, opened 2017-01-23
Stream and talk endpoints

ChrisGuzman:


src/main/java/com/nexmo/client/voice/StreamRequest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.nexmo.client.NexmoUnexpectedException;
+
+public class StreamRequest {
+    private StreamPayload streamPayload;
+
+    public StreamRequest(String streamUrl, boolean loop) {
+        this.streamPayload = new StreamPayload(streamUrl, (loop ? 1 : 0));
```
ChrisGuzman: Should be `(loop ? 0 : 1)` 

0 means loop. 1 means don't loop.
The default is not to loop.

https://docs.nexmo.com/voice/voice-api/api-reference#cspparameters

ChrisGuzman:
Let's add some model tests and make sure that we have a test that ensures we're looping when we should be and not looping when we shouldn't be

src/main/java/com/nexmo/client/voice/endpoints/StartStreamMethod.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.util.EntityUtils;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+
+public class StartStreamMethod extends AbstractMethod<StreamRequest, NexmoResponse> {
+    private static final Log LOG = LogFactory.getLog(CreateCallMethod.class);
```
ChrisGuzman: Should be `LogFactory.getLog(StartStreamMethod.class);`

src/main/java/com/nexmo/client/voice/StreamRequest.java
judy2k: Looking at the stream NCCO, I think 0 means "loop forever", other values mean "Play _n_ times", so I'm not sure hiding this value behind a boolean makes sense.

ChrisGuzman:
@judy2k This is ready for review and eventual merging. I think some tests can be added but since we're above 80% I think tests can be added in another PR. What do you think?

judy2k:
Sounds good.

src/main/java/com/nexmo/client/voice/StopTalkMethod.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    protected Class[] getAcceptableAuthMethods() {
+        return ALLOWED_AUTH_METHODS;
+    }
+
+    @Override
+    public HttpUriRequest makeRequest(String uuid) throws NexmoClientException, UnsupportedEncodingException {
+        String uri = this.uri + uuid + "/talk";
+        HttpDelete delete = new HttpDelete(uri);
+        delete.setHeader("Content-Type", "application/json");
```
judy2k: Do we need this if there's no body to the DELETE request?

src/main/java/com/nexmo/client/DTMFEndpoint.java
```diff
@@ -37,7 +35,7 @@ public DTMFEndpoint(HttpWrapper httpWrapper) {
         this.sendDTMF = new SendDTMFMethod(httpWrapper);
     }
 
-    public DTMFResponse put(String uuid, String digits) throws IOException, NexmoClientException {
+    public NexmoResponse put(String uuid, String digits) throws IOException, NexmoClientException {
```
ChrisGuzman: Yeah for Stream, Talk, and DTMF, the response consists of a `message` and a `uuid` the only thing that differs is that the uuid will be unique (duh) and that the message will be "(Stream/Talk/DTMF) (Started/Stopped)"

src/main/java/com/nexmo/client/voice/StopTalkMethod.java
ChrisGuzman: The code snippets in the VAPI docs include adding the header 
https://docs.nexmo.com/voice/voice-api/api-reference#talk_delete
https://docs.nexmo.com/voice/voice-api/api-reference#stream_delete

src/main/java/com/nexmo/client/voice/NexmoResponse.java
```diff
@@ -25,24 +25,25 @@
 
 import java.io.IOException;
 
-public class DTMFResponse {
-    private String message;
+public class NexmoResponse {
```
ChrisGuzman: Will do

src/main/java/com/nexmo/client/voice/StopTalkMethod.java
judy2k: It's not going to do any harm, but I'd argue it's incorrect. Happy to leave it.

judy2k:
:shipit:

judy2k:
I'll have a look tomorrow if you like. Getting late here now!

ChrisGuzman:
Thanks!

src/test/java/com/nexmo/client/voice/endpoints/CallRecordTest.java
```diff
@@ -73,13 +74,35 @@ public void testJson() throws Exception {
...
+        Calendar expectedStart = new GregorianCalendar(2017, Calendar.JANUARY, 13 ,13, 55, 2);
+        expectedStart.getTime();
+        expectedStart.setTimeZone(TimeZone.getTimeZone("UTC"));
+
+        Calendar actualStart = Calendar.getInstance();
+        actualStart.setTimeZone(TimeZone.getTimeZone("UTC"));
+        actualStart.setTime(record.getStartTime());
+
         assertEquals(
-                new GregorianCalendar(2017, Calendar.JANUARY, 13 ,13, 55, 2).getTime(),
```
ChrisGuzman: by calling `getTime()` the test defaults to the current local timezone. Thus on my machine this is going to equal 13:55:2 EST.

ChrisGuzman:
The problem with debugging this test is that when a test fails it prints out the Date as a string. Printing out a date as a string converts that time into your local timezone so even if you set the timezone correctly for a Date or Calendar, you're still going to get the wrong timezone in the test result

src/test/java/com/nexmo/client/voice/endpoints/CallRecordTest.java
```diff
@@ -73,13 +74,35 @@ public void testJson() throws Exception {
...
-                record.getStartTime());
```
ChrisGuzman: Then calling `getStartTime()` is going to take the json string in UTC timezone, then convert it to EST. Thus it's going to print out as 8:55:2

PR #26, opened 2017-01-27
Ncco servlet

judy2k:
Adds a servlet for implementing Ncco webhooks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ apply plugin: 'java'
 apply plugin: "jacoco"
 apply plugin: 'maven'
 apply plugin: 'eclipse'
-apply plugin: 'jetty'
+apply plugin: 'war'
 
 group = "com.nexmo"
 version = "2.0.0-prerelease"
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -13,130 +13,140 @@ apply plugin: 'java'
 apply plugin: "jacoco"
 apply plugin: 'maven'
 apply plugin: 'eclipse'
+apply plugin: 'war'
```
ChrisGuzman: huh. What is it good for?

src/main/java/com/nexmo/client/voice/servlet/AbstractAnswerServlet.java
```diff
@@ -0,0 +1,51 @@
...
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public abstract class AbstractAnswerServlet extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
```
ChrisGuzman: why `doGet()` instead of `get()`

judy2k: Because the servlet specification requires `doGet` (it's overriding a method in the superclass)

build.gradle
judy2k: `gradle war` ➡️  builds a war file containing the example servlet that can then be deployed to Jetty or Tomcat to test.

src/main/java/com/nexmo/client/voice/servlet/AbstractAnswerServlet.java
ChrisGuzman: Oh duh. Missed that @Override.

PR #27, opened 2017-01-31
Abstract method interface change

judy2k:
Needed to replace `HttpUriRequest` with `RequestBuilder` in AbstractMethod because other auth methods need to change things other than just headers (and HttpUriRequest is almost immutable).

### DIFF
--- a/src/main/java/com/nexmo/client/voice/StartTalkMethod.java
+++ b/src/main/java/com/nexmo/client/voice/StartTalkMethod.java
@@ -27,8 +27,7 @@ import com.nexmo.client.voice.endpoints.AbstractMethod;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.HttpResponse;
-import org.apache.http.client.methods.HttpPut;
-import org.apache.http.client.methods.HttpUriRequest;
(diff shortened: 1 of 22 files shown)


src/main/java/com/nexmo/client/voice/StartTalkMethod.java
```diff
@@ -52,14 +51,11 @@ public StartTalkMethod(HttpWrapper httpWrapper) {
     }
 
     @Override
-    public HttpUriRequest makeRequest(TalkRequest request) throws NexmoClientException, UnsupportedEncodingException {
+    public RequestBuilder makeRequest(TalkRequest request) throws NexmoClientException, UnsupportedEncodingException {
         String uri = this.uri + request.getUuid() + "/talk";
-        HttpPut put = new HttpPut(uri);
-        put.setHeader("Content-Type", "application/json");
-        put.setEntity(new StringEntity(request.toJson()));
-        LOG.info("Request: " + request.toJson());
```
ChrisGuzman: Are we getting rid of the LOG statements?

judy2k: My plan is actually to put log statements in consistent places in the code. But given we haven't worked out where in the code are useful places to log values, I figure it's no harm to remove these right now.

PR #28, opened 2017-01-31
Sip and websocket endpoints

ChrisGuzman:


ChrisGuzman:
We might want to change some of the getters/setters/constructors since now there are three endpoints that a Call or CallRecord can have. I figured you and I could have a chat about that before I start changing those.

Not sure why I'm missing the code coverage report.

src/main/java/com/nexmo/client/voice/CallRecord.java
```diff
@@ -62,19 +62,29 @@ public CallRecord(CallEndpoint to, CallEndpoint from) {
         this.from = from;
     }
 
-    public CallEndpoint getTo() {
+    public CallRecord(SipEndpoint to, SipEndpoint from) {
```
judy2k: Why don't you just have `public CallRecord(Endpoint to, Endpoint from)` ?

src/main/java/com/nexmo/client/voice/Endpoint.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.EXISTING_PROPERTY,
+        property = "type"
+)
+@JsonSubTypes({
```
judy2k: Great! I knew there was a way of doing this, but I couldn't find the docs for it again.

src/main/java/com/nexmo/client/voice/SipEndpoint.java
```diff
@@ -0,0 +1,51 @@
...
+        return uri;
+    }
+
+    @Override
+    public String getType() {
+        return type;
+    }
+
+    @Override
+    public String toLog() {
```
judy2k: Would it be worth using `toString` for this? Happy to leave it though.

src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -36,8 +36,8 @@
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties({ "_links"})
 public class Call {
-    private CallEndpoint to;
-    private CallEndpoint from;
+    private Endpoint to;
```
judy2k: Ugh, I wish endpoint didn't mean (a) the origin/destination of a phone call; (b) a REST endpoint. Naming is hard.

src/main/java/com/nexmo/client/voice/CallEndpoint.java
```diff
@@ -25,7 +25,7 @@
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
-public class CallEndpoint {
+public class CallEndpoint implements Endpoint {
```
judy2k: I think this should be called "PhoneEndpoint" to match its `type`. What do you think?

src/main/java/com/nexmo/client/voice/CallRecord.java
ChrisGuzman: I thought it might be helpful to the users to spell out that there are three different types of endpoints a CallRecord can be. But then again since `CallRecord` is just for showing you details about a call it's useless here.

src/main/java/com/nexmo/client/voice/SipEndpoint.java
ChrisGuzman: I went back and forth on this in my head. I was afraid that overriding toString might mess with some logging that users might want to do. I have no strong feelings about `toString()` or `toLog()`

src/main/java/com/nexmo/client/voice/CallEndpoint.java
ChrisGuzman: Agreed

judy2k:
I think the Endpoint interface does that, and it's kinda what interfaces
are for.

PR #31, opened 2017-02-02
Use ModifyCallResponse

ChrisGuzman:
The API returns that, not a Call type object

StreamPayload takes streamUrl as a list

Update Call object

### DIFF
--- a/src/main/java/com/nexmo/client/voice/Call.java
+++ b/src/main/java/com/nexmo/client/voice/Call.java
@@ -36,7 +36,7 @@ import java.io.IOException;
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties({ "_links"})
 public class Call {
-    private Endpoint to;
+    private Endpoint[] to;
     private Endpoint from;
     private String answerUrl;
 
@@ -53,30 +53,18 @@ public class Call {
         this(new PhoneEndpoint(to), new PhoneEndpoint(from), answerUrl);
     }
 
-    public Call(PhoneEndpoint to, PhoneEndpoint from, String answerUrl) {
-        this.to = to;
-        this.from = from;
-        this.answerUrl = answerUrl;
-    }
-
-    public Call(SipEndpoint to, SipEndpoint from, String answerUrl) {
-        this.to = to;
-        this.from = from;
-        this.answerUrl = answerUrl;
-    }
-
-    public Call(WebSocketEndpoint to, WebSocketEndpoint from, String answerUrl) {
-        this.to = to;
+    public Call(Endpoint to, Endpoint from, String answerUrl) {
+        this.to = new Endpoint[]{to};
         this.from = from;
         this.answerUrl = answerUrl;
     }
 
     public Endpoint[] getTo() {
-        return new Endpoint[]{to};
+        return to;
     }
 
     public void setTo(Endpoint[] to) {
-        this.to = to[0];
+        this.to = to;
     }
 
     public Endpoint getFrom() {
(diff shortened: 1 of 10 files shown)


src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -53,30 +53,18 @@ public Call(String to, String from, String answerUrl) {
...
-
-    public Call(SipEndpoint to, SipEndpoint from, String answerUrl) {
-        this.to = to;
-        this.from = from;
-        this.answerUrl = answerUrl;
-    }
-
-    public Call(WebSocketEndpoint to, WebSocketEndpoint from, String answerUrl) {
-        this.to = to;
+    public Call(Endpoint to, Endpoint from, String answerUrl) {
```
judy2k: There should probably be another constructor that takes `Endpoint[] to` so the user can provide more than one endpoint if they wish to.

src/main/java/com/nexmo/client/voice/NexmoVoiceClient.java
```diff
@@ -66,7 +66,7 @@ public DtmfResponse sendDtmf(String uuid, String digits) throws IOException, Nex
         return dtmf.put(uuid, digits);
     }
 
-    public CallRecord modifyCall(String uuid, String action) throws IOException, NexmoClientException {
+    public ModifyCallResponse modifyCall(String uuid, String action) throws IOException, NexmoClientException {
```
judy2k: Glad we caught this before release!

src/main/java/com/nexmo/client/voice/StreamPayload.java
```diff
@@ -39,7 +39,7 @@ public int getLoop() {
     }
 
     public StreamPayload(String streamUrl, int loop) {
```
judy2k: Even though it's provided as an array, does the API only support a single streamUrl at the moment? If so, leave this, if it does actually support more than one streamUrl, there should probably be another constructor that takes `String[] streamUrl` as a parameter.

src/test/java/com/nexmo/client/voice/NexmoVoiceClientTest.java
```diff
@@ -177,34 +177,9 @@ public void testSendDtmf() throws Exception {
...
-                "        \"_links\": {\n" +
-                "          \"self\": {\n" +
-                "            \"href\": \"/v1/calls/93137ee3-580e-45f7-a61a-e0b5716000ef\"\n" +
-                "          }\n" +
-                "        }\n" +
-                "      }\n"));
-        CallRecord call = client.modifyCall("93137ee3-580e-45f7-a61a-e0b5716000ef", "hangup");
-        assertEquals("93137ee3-580e-45f7-a61a-e0b5716000ef", call.getUuid());
-
+        NexmoVoiceClient client = new NexmoVoiceClient(stubHttpWrapper(200, "{\"message\":\"Received\"}"));
```
judy2k: Well that JSON's a bit easier ;-)

src/test/java/com/nexmo/client/voice/StreamPayloadTest.java
```diff
@@ -0,0 +1,54 @@
+package com.nexmo.client.voice;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/*
```
judy2k: How did your copyright end up down here?!

src/main/java/com/nexmo/client/voice/ModifyCallResponse.java
```diff
@@ -0,0 +1,46 @@
+package com.nexmo.client.voice;/*
```
judy2k: Copyright should be above the package declaration.

src/main/java/com/nexmo/client/voice/StreamPayload.java
ChrisGuzman: I talked to @sammachin yesterday and he said that currently the API only supports taking a single `streamUrl`, but it must be in an array

judy2k: OK

PR #33, opened 2017-02-03
Add some javadoc comments to DTMF, Stream, and Talk objects

ChrisGuzman:
@judy2k can you give me some feedback on these?

### DIFF
--- a/src/main/java/com/nexmo/client/voice/TalkPayload.java
+++ b/src/main/java/com/nexmo/client/voice/TalkPayload.java
@@ -23,6 +23,15 @@ package com.nexmo.client.voice;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+/**
+ * The JSON payload that will be sent in a {@link TalkRequest}.
+ * <p>
+ * {@code text}: A UTF-8 and URL encoded string of up to 1500 characters containing the message to be synthesized
+ * in the Call or Conversation. Each comma in text adds a short pause to the synthesized speech.
+ * {@link VoiceName}: The name of the voice used to deliver {@code text}.
+ * {@code loop}: The number of times the audio file at stream_url is repeated before the stream ends.
+ */
+
 public class TalkPayload {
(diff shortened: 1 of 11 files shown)


src/main/java/com/nexmo/client/voice/TalkPayload.java
```diff
@@ -23,6 +23,15 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+/**
+ * The JSON payload that will be sent in a {@link TalkRequest}.
+ * <p>
+ * {@code text}: A UTF-8 and URL encoded string of up to 1500 characters containing the message to be synthesized
```
judy2k: The user doesn't need to know anything about UTF-8 or URL encoding - they just fill the String and the encoding is done by the Jackson, no?

src/main/java/com/nexmo/client/voice/TalkPayload.java
```diff
@@ -23,6 +23,15 @@
...
+ * in the Call or Conversation. Each comma in text adds a short pause to the synthesized speech.
+ * {@link VoiceName}: The name of the voice used to deliver {@code text}.
+ * {@code loop}: The number of times the audio file at stream_url is repeated before the stream ends.
```
judy2k: Specify that loop should be 0 for infinite looping? (_is_ that the case?)

src/main/java/com/nexmo/client/voice/TalkPayload.java
ChrisGuzman: true. I took some of that from the docs, so that part is unnecessary.

src/main/java/com/nexmo/client/voice/TalkPayload.java
ChrisGuzman: Good point

ChrisGuzman:
@judy2k can you review and merge when you have a chance?

PR #45, opened 2017-03-02
Resolve a couple of dependency conflicts.

judy2k:
It turns out when two conflicting versions of a dependency are required, Gradle just picks the latest one. You can see this when there's an arrow `->` in the output from `gradle dependencies`.

I updated our direct dependencies to use the version that was being used anyway, thus resolving what was technically a version conflict.

I also added a new `zip` task to the gradle build file, that creates a zip file containing the client library and jar files for all the dependencies.

## Contibution Checklist
* [x] Unit tests!
* [x] Updated [CHANGELOG.md]
* [x] My name is in [CONTRIBUTORS.md]


judy2k:
I checked "updated CHANGELOG.md" because this change to the build file doesn't affect the produced library, and so shouldn't affect any users.

PR #57, opened 2017-03-29
Implement cnam parameter for Number Insight

judy2k:
Added the new `cnam` parameter for Standard and Advanced Insight requests, along with the necessary additions to the response objects.

## Contribution Checklist
* [x] Unit tests!
* [x] Updated [CHANGELOG.md]
* [x] My name is in [CONTRIBUTORS.md]

### DIFF
--- a/src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
+++ b/src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
@@ -25,19 +25,32 @@ import com.nexmo.client.insight.BaseInsightRequest;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 
 public class AdvancedInsightRequest extends BaseInsightRequest {
-    private String ipAddress;
+    private static final Boolean DEFAULT_CNAM = null;
+    private final static String DEFAULT_IP_ADDRESS = null;
+    private final String ipAddress;
+    private final Boolean cnam;
+
 
     public AdvancedInsightRequest(String number) {
-        super(number, null);
+        this(number, DEFAULT_COUNTRY);
     }
 
     public AdvancedInsightRequest(String number, String country) {
-        super(number, country);
+        this(number, country, DEFAULT_IP_ADDRESS);
     }
 
     public AdvancedInsightRequest(String number, String country, String ipAddress) {
+        this(number, country, ipAddress, DEFAULT_CNAM);
+    }
+
+    public AdvancedInsightRequest(String number, String country, String ipAddress, Boolean cnam) {
         super(number, country);
+        this.cnam = cnam;
         this.ipAddress = ipAddress;
+    }
+
+    public Boolean getCnam() {
+        return cnam;
     }
 
     public String getIpAddress() {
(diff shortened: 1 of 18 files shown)


src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
```diff
@@ -25,21 +25,34 @@
 import org.apache.commons.lang3.builder.EqualsBuilder;
 
 public class AdvancedInsightRequest extends BaseInsightRequest {
-    private String ipAddress;
+    private static final Boolean DEFAULT_CNAM = null;
+    private final static String DEFAULT_IP_ADDRESS = null;
```
ChrisGuzman: Can we agree on the order of `private static final` vs `private final static`

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
```diff
@@ -25,21 +25,34 @@
...
+
 
     public AdvancedInsightRequest(String number) {
-        super(number, null);
+        this(number, DEFAULT_COUNTRY);
     }
 
     public AdvancedInsightRequest(String number, String country) {
-        super(number, country);
+        this(number, country, DEFAULT_IP_ADDRESS);
```
ChrisGuzman: why use a constant initialized to `null` instead of passing in `null`?

ChrisGuzman:
Also looks like codacy wants you to add some assertions to your tests, check that out.

judy2k:
Codacy wants me to add some assertions to my tests, but only because it doesn't understand the assertions used by Mockito.

judy2k:
I guess we may want to add some extra info about charges related to cnam, but given that we have no javadocs for this yet, I think that can be left until then. The user will be reading docs.nexmo.com anyway to figure these params out.

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
judy2k: I don't know if we can agree - but it should be `public static final` ;-)

I think the various 'standards' agree on this, and it's implemented by [checkstyle](http://checkstyle.sourceforge.net/config_modifier.html)

judy2k: Really not sure how I got this wrong, actually. Must be to do with the way I implemented it.

ChrisGuzman: Should they be public or private?

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightRequest.java
judy2k: I fought with this a bit while developing it, and the main reason is that if the default value of ipAddress was to change, we would only have to change it in the one place, rather than attempting to find everywhere `null` was being passed.

I think this version, with the static value is probably more correct, but agree that it's not quite idiomatic. If it's okay, I'd like to keep it for now as it doesn't really affect the user, and I'd like to see which I prefer in the long run.

ChrisGuzman: I'm fine with keeping it. I meant to ask more as a question instead of an intent to change it back

ChrisGuzman:
@judy2k feel free to merge when ready

PR #64, opened 2017-04-23
Send Event Url as a list

### LINKED ISSUE #63: using setEventUrl for Voice Call returns an invalid parameter Bad Request
When calling setEventUrl on a Call object, and then calling the VoiceClient createCall function using this object, the returned CallEvent has no set objects and no call is made. Output shows:

Received: {"type":400,"title":"Bad Request","invalid_parameters":[{"reason":"Invalid type parameter provided","name":"event_url"}]}

If setEventUrl is not called, then the CallEvent is correctly set and a call is made.

ChrisGuzman:
Fixes #63

Treat `eventUrl` as an array, not as a String

## Contribution Checklist
* [X] Unit tests!
* [X] My name is in [CONTRIBUTORS.md](CONTRIBUTORS.md)

### DIFF
--- a/src/main/java/com/nexmo/client/voice/Call.java
+++ b/src/main/java/com/nexmo/client/voice/Call.java
@@ -101,8 +101,11 @@ public class Call {
     }
 
     @JsonProperty("event_url")
-    public String getEventUrl() {
-        return eventUrl;
+    public String[] getEventUrl() {
+        if (eventUrl == null) {
+            return null;
+        }
+        return new String[]{eventUrl};
     }
 
     public void setEventUrl(String eventUrl) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -101,8 +101,11 @@ public void setAnswerMethod(String answerMethod) {
     }
 
     @JsonProperty("event_url")
-    public String getEventUrl() {
-        return eventUrl;
+    public String[] getEventUrl() {
+        if (eventUrl == null) {
+            return null;
```
ChrisGuzman: Without this the json generated for a call with a null event url would be `event_url = [null]`

PR #77, opened 2017-08-07
Allow users to change the base URI for requests

### LINKED ISSUE #58: Should allow baseUrl to be changed for all client objects

ChrisGuzman:
Allow users to change the base URI for requests. Fixes #58

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/src/main/java/com/nexmo/client/NexmoClient.java
+++ b/src/main/java/com/nexmo/client/NexmoClient.java
@@ -120,6 +120,7 @@ public class NexmoClient {
      */
     public void setBaseUri(String baseUri) {
         this.baseUri = baseUri;
+
         this.insight.setBaseUri(this.baseUri);
         this.verify.setBaseUri(this.baseUri);
         this.voice.setBaseUri(this.baseUri);
--- a/src/main/java/com/nexmo/client/NexmoClient.java
+++ b/src/main/java/com/nexmo/client/NexmoClient.java
@@ -103,6 +103,13 @@ public class NexmoClient {
         return this.voice;
     }
 
+    /**
+     * The server this library calls when making requests. Will be null if no overriding value has been set.
+     * If this is null the client's {@code getBaseUri} method should be called.
+     *
+     * @return String uri the library calls when making requests or null
+     */
+
     public String getBaseUri() {
         return baseUri;
     }
@@ -113,5 +120,10 @@ public class NexmoClient {
      */
     public void setBaseUri(String baseUri) {
         this.baseUri = baseUri;
+        this.insight.setBaseUri(this.baseUri);
+        this.verify.setBaseUri(this.baseUri);
+        this.voice.setBaseUri(this.baseUri);
+        this.sms.setBaseUri(this.baseUri);
+        this.sns.setBaseUri(this.baseUri);
     }
 }
--- a/src/main/java/com/nexmo/client/NexmoClient.java
+++ b/src/main/java/com/nexmo/client/NexmoClient.java
@@ -46,7 +46,20 @@ public class NexmoClient {
     private final VerifyClient verify;
     private final SnsClient sns;
 
+    private String baseUri;
     private HttpWrapper httpWrapper;
+
+
+    public NexmoClient(String baseUri, AuthMethod... authMethods) {
+        this.baseUri = baseUri;
+        this.httpWrapper = new HttpWrapper(authMethods);
+
+        this.insight = new InsightClient(this.httpWrapper, baseUri);
+        this.verify = new VerifyClient(this.httpWrapper, baseUri);
+        this.voice = new VoiceClient(this.httpWrapper, baseUri);
+        this.sms = new SmsClient(this.httpWrapper, baseUri);
+        this.sns = new SnsClient(this.httpWrapper, baseUri);
+    }
 
     public NexmoClient(AuthMethod... authMethods) {
         this.httpWrapper = new HttpWrapper(authMethods);
@@ -82,4 +95,11 @@ public class NexmoClient {
         return this.voice;
     }
 
+    public String getBaseUri() {
+        return baseUri;
+    }
+
+    public void setBaseUri(String baseUri) {
+        this.baseUri = baseUri;
+    }
 }
--- a/src/main/java/com/nexmo/client/NexmoClient.java
+++ b/src/main/java/com/nexmo/client/NexmoClient.java
@@ -36,8 +36,6 @@ import org.apache.http.client.HttpClient;
  * Construct an instance of this object with one or more {@link AuthMethod}s (providing all the authentication methods
  * for the APIs you wish to use), and then call {@link #getVoiceClient()} to obtain a client for the Nexmo Voice API.
  * <p>
- * Currently this object only constructs and provides access to {@link VoiceClient}. In the future it will manage
- * clients for all of the Nexmo APIs.
  */
 public class NexmoClient {
     private final InsightClient insight;
@@ -49,7 +47,12 @@ public class NexmoClient {
     private String baseUri;
     private HttpWrapper httpWrapper;
 
-
+    /**
+     *
+     * @param baseUri Change the server this library calls when making requests. <b>Don't include a trailing slash</b>
+     * @param authMethods (required) one or more {@link AuthMethod}s (providing all the authentication methods
+     * for the APIs you wish to use)
+     */
     public NexmoClient(String baseUri, AuthMethod... authMethods) {
         this.baseUri = baseUri;
         this.httpWrapper = new HttpWrapper(authMethods);
@@ -61,6 +64,11 @@ public class NexmoClient {
         this.sns = new SnsClient(this.httpWrapper, baseUri);
     }
 
+    /**
+     *
+     * @param authMethods (required) one or more {@link AuthMethod}s (providing all the authentication methods
+     * for the APIs you wish to use)
+     */
     public NexmoClient(AuthMethod... authMethods) {
         this.httpWrapper = new HttpWrapper(authMethods);
 
@@ -99,6 +107,10 @@ public class NexmoClient {
         return baseUri;
     }
 
+    /**
+     *
+     * @param baseUri Change the server this library calls when making requests. <b>Don't include a trailing slash</b>
+     */
     public void setBaseUri(String baseUri) {
         this.baseUri = baseUri;
     }
--- a/src/main/java/com/nexmo/client/insight/InsightClient.java
+++ b/src/main/java/com/nexmo/client/insight/InsightClient.java
@@ -112,5 +112,9 @@ public class InsightClient {
 
     public void setBaseUri(String baseUri) {
         this.baseUri = baseUri;
+
+        this.basic.setBaseUri(baseUri);
+        this.standard.setBaseUri(baseUri);
+        this.advanced.setBaseUri(baseUri);
     }
 }
--- a/src/main/java/com/nexmo/client/insight/InsightClient.java
+++ b/src/main/java/com/nexmo/client/insight/InsightClient.java
@@ -103,6 +103,4 @@ public class InsightClient {
             String number, String country, String ipAddress, boolean cnam) throws IOException, NexmoClientException {
         return this.advanced.execute(new AdvancedInsightRequest(number, country, ipAddress, cnam));
     }
-
-
 }
--- a/src/main/java/com/nexmo/client/insight/InsightClient.java
+++ b/src/main/java/com/nexmo/client/insight/InsightClient.java
@@ -44,6 +44,7 @@ public class InsightClient {
     protected BasicInsightEndpoint basic;
     protected StandardInsightEndpoint standard;
     protected AdvancedInsightEndpoint advanced;
+    private String baseUri;
 
     /**
      * Constructor.
@@ -57,6 +58,7 @@ public class InsightClient {
     }
 
     public InsightClient(HttpWrapper httpWrapper, String baseUri) {
+        this.baseUri = baseUri;
         this.basic = new BasicInsightEndpoint(httpWrapper, baseUri);
         this.standard = new StandardInsightEndpoint(httpWrapper, baseUri);
         this.advanced = new AdvancedInsightEndpoint(httpWrapper, baseUri);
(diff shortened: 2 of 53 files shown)


ChrisGuzman:
Do we want a trailing slash on a url or not?

https://github.com/Nexmo/nexmo-java/blob/master/src/main/java/com/nexmo/client/voice/endpoints/CreateCallMethod.java#L42
vs
https://github.com/Nexmo/nexmo-java/blob/master/src/main/java/com/nexmo/client/voice/endpoints/ReadCallMethod.java#L39

ChrisGuzman:
Let's test this IRL

ChrisGuzman:
Hi @stefanscheidt we noticed you thumbs up'ed #58, the issue that this PR aims to resolve. Would you like to look over this PR and let us know your thoughts?

ChrisGuzman:
@judy2k it looks like you had the idea a while back about changing the baseUri for a request https://github.com/Nexmo/nexmo-java/blob/master/src/main/java/com/nexmo/client/voice/endpoints/CreateCallMethod.java#L70

https://github.com/Nexmo/nexmo-java/blob/master/src/test/java/com/nexmo/client/voice/endpoints/CreateCallMethodTest.java#L92

src/main/java/com/nexmo/client/insight/InsightClient.java
```diff
@@ -97,4 +105,12 @@ public AdvancedInsightResponse getAdvancedNumberInsight(
             String number, String country, String ipAddress, boolean cnam) throws IOException, NexmoClientException {
         return this.advanced.execute(new AdvancedInsightRequest(number, country, ipAddress, cnam));
     }
+
+    public String getBaseUri() {
+        return baseUri;
+    }
+
+    public void setBaseUri(String baseUri) {
```
judy2k: Shouldn't this change the baseUri on all the Endpoint objects?

judy2k:
I knew we wanted to be able to change the baseUri when this stuff was originally developed, but didn't have time before the 2.0 release to implement it.

src/main/java/com/nexmo/client/voice/endpoints/StopStreamMethod.java
```diff
@@ -37,14 +37,19 @@
...
     private static final Class[] ALLOWED_AUTH_METHODS = new Class[]{JWTAuthMethod.class};
     private String uri = DEFAULT_URI;
 
     public StopStreamMethod(HttpWrapper httpWrapper) {
         super(httpWrapper);
     }
 
+    public StopStreamMethod(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + "/v1/calls";
```
judy2k: I think my general preference is to *have* the slash at the end, but I'm not sure whether there's one version that's more correct, or even more of a standard internally. So as long as we're consistent I'm happy and I'll go with whatever you're happy with.

ChrisGuzman: I decided to remove the slash at the end since that's the way the library had it before we started making changes and the docs all omit a trailing slash https://developer.nexmo.com/api/voice

src/main/java/com/nexmo/client/NexmoClient.java
```diff
@@ -82,4 +95,11 @@ public VoiceClient getVoiceClient() {
         return this.voice;
     }
 
+    public String getBaseUri() {
```
ChrisGuzman: Done

src/main/java/com/nexmo/client/insight/InsightClient.java
ChrisGuzman: Good call. Change made and test updated

src/main/java/com/nexmo/client/NexmoClient.java
```diff
@@ -36,8 +36,6 @@
  * Construct an instance of this object with one or more {@link AuthMethod}s (providing all the authentication methods
  * for the APIs you wish to use), and then call {@link #getVoiceClient()} to obtain a client for the Nexmo Voice API.
  * <p>
- * Currently this object only constructs and provides access to {@link VoiceClient}. In the future it will manage
```
ChrisGuzman: This isn't true anymore right @judy2k

src/test/java/com/nexmo/client/insight/InsightClientTest.java
```diff
@@ -99,4 +101,21 @@ public void testGetAdvancedNumberInsight3() throws Exception {
         verify(client.advanced).execute(eq(new AdvancedInsightRequest("12345", "GB", "123.123.123.123", true)));
     }
 
+    @Test
+    public void testSetBaseUri() throws Exception {
+        assertNull(this.client.getBaseUri());
+        this.client.setBaseUri("https://example.com");
+        assertEquals("https://example.com", this.client.getBaseUri());
+
+        assertEquals("https://example.com/ni/basic/json", this.client.basic.getUri());
```
ChrisGuzman: These tests fail because we're mocking the clients. Can you help me out with this @judy2k?

ChrisGuzman: Nevermind. Just adjusted the test a bit.

src/main/java/com/nexmo/client/insight/basic/BasicInsightEndpoint.java
```diff
@@ -45,6 +45,20 @@ public BasicInsightEndpoint(HttpWrapper httpWrapper) {
         super(httpWrapper);
     }
 
+    public BasicInsightEndpoint(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + "/ni/basic/json";
```
stefanscheidt: Probably extract "/ni/basic/json" in a constant like INSIGHT_ENDPOINT_PATH and also use it in setBaseUri?

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightEndpoint.java
```diff
@@ -44,6 +44,19 @@ public AdvancedInsightEndpoint(HttpWrapper httpWrapper) {
         super(httpWrapper);
     }
 
+    public AdvancedInsightEndpoint(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + "/ni/advanced/json";
```
stefanscheidt: Probably extract "/ni/advanced/json" in a constant like ADVANCED_ENDPOINT_PATH and also use it in setBaseUri?

src/main/java/com/nexmo/client/sms/SendMessageEndpoint.java
```diff
@@ -57,6 +57,15 @@ public SendMessageEndpoint(HttpWrapper httpWrapper) {
...
 
+    public SendMessageEndpoint(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + "/sms/xml";
+    }
+
+    public String getUri() {
+        return uri;
+    }
+
```
stefanscheidt: No setBaseUri here?

src/main/java/com/nexmo/client/sms/SmsClient.java
```diff
@@ -65,4 +71,11 @@ public SmsClient(HttpWrapper httpWrapper) {
...
     }
 
-}
+     public String getBaseUri() {
+         return baseUri;
+     }
+
+     public void setBaseUri(String baseUri) {
+         this.baseUri = baseUri;
+     }
```
stefanscheidt: Should a this.message.setBaseUri(baseUri) be called here?

src/main/java/com/nexmo/client/sns/SnsClient.java
```diff
@@ -48,11 +49,25 @@ public SnsClient(HttpWrapper httpWrapper) {
...
         return (SnsSubscribeResponse) this.endpoint.execute(request);
     }
+
+    public String getBaseUri() {
+        return baseUri;
+    }
+
+    public void setBaseUri(String baseUri) {
+        this.baseUri = baseUri;
+    }
```
stefanscheidt: add this.endpoint.setBaseUri(baseUri) ?

src/main/java/com/nexmo/client/sns/SnsEndpoint.java
```diff
@@ -60,6 +60,15 @@ public SnsEndpoint(HttpWrapper httpWrapper) {
...
 
+    public SnsEndpoint(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + "/sns/xml";
+    }
+
+    public String getUri() {
+        return uri;
+    }
+
```
stefanscheidt: add setBaseUri?

src/test/java/com/nexmo/client/NexmoClientTest.java
```diff
@@ -76,4 +76,49 @@ public void testConstructNexmoClient() throws Exception {
...
+                "  \"direction\": \"outbound\"\n" +
+                "}"));
+        assertEquals("https://example.com", client.getBaseUri());
+        CallEvent evt = client.getVoiceClient().createCall(new Call("4499991111", "44111222333", "https://callback.example.com/"));
+        assertEquals(CallStatus.STARTED, evt.getStatus());
+    }
+
+
+    @Test
+    public void testBseUriSetterOnNexmoClient() throws Exception {
```
stefanscheidt: Typo: testBaseUri...

src/main/java/com/nexmo/client/sms/SmsClient.java
judy2k: Yeah, @ChrisGuzman, you need to go through each of the XxxClient classes and ensure that `setBaseUri` propagates its value to the Endpoints contained within each client.

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightEndpoint.java
judy2k: I agree with @stefanscheidt . Sorry for setting a bad precedent here, @ChrisGuzman.

I'm happy for this to be done in a separate PR, given that we'd really like to get this released soon.

ChrisGuzman:
@judy2k wanna give this another review?

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightEndpoint.java
```diff
@@ -36,14 +36,28 @@
...
     private static final Class[] ALLOWED_AUTH_METHODS = new Class[]{SignatureAuthMethod.class, TokenAuthMethod.class};
 
-    private static final String DEFAULT_URI = "https://api.nexmo.com/ni/advanced/json";
+    private static final String DEFAULT_URI  = "https://api.nexmo.com/ni/advanced/json";
+    private static final String DEFAULT_PATH = "/ni/advanced/json";
 
     private String uri = DEFAULT_URI;
 
     public AdvancedInsightEndpoint(HttpWrapper httpWrapper) {
         super(httpWrapper);
```
judy2k: I always think it's a good idea to funnel all constructor calls through a single "master" constructor. That way, if you want to change the way a class is constructed there's only one place you need to make the change, in theory.

Given this, what do you think of changing `super(httpWrapper)` here to `this(httpWrapper, DEFAULT_BASE_URI)`, and removing DEFAULT_URI?

src/main/java/com/nexmo/client/sms/SmsClient.java
```diff
@@ -65,4 +71,12 @@ public SmsClient(HttpWrapper httpWrapper) {
         return this.message.execute(message);
     }
 
-}
+     public String getBaseUri() {
```
judy2k: If I construct this with `SmsClient(httpWrapper)` and then call `getBaseUri()`, instead of returning the default, it's going to return `null` and that bugs me a little bit.

src/main/java/com/nexmo/client/sns/SnsEndpoint.java
```diff
@@ -60,6 +62,19 @@ public SnsEndpoint(HttpWrapper httpWrapper) {
...
+    public SnsEndpoint(HttpWrapper httpWrapper, String baseUri) {
+        super(httpWrapper);
+        uri = baseUri + DEFAULT_PATH;
+    }
+
+    public String getUri() {
+        return uri;
+    }
+
+    public void setUri(String uri) {
```
judy2k: This should really be `setBaseUri()`

src/main/java/com/nexmo/client/sms/SmsClient.java
cr0wst: I know this is an old comment but I think that can be solved similar to your comment about constructors calling through a single master constructor.  The `SmsClient(HttpWrapper)` constructor should call the `SmsClient(HttpWrapper, string)` constructor with the default URI.

src/main/java/com/nexmo/client/insight/advanced/AdvancedInsightEndpoint.java
cr0wst: Again, old comment, but I wanted to chime in that I agree with this.  This should, ideally, also be the common practice for overloaded methods.  Starting with the lowest number of arguments and working your way in.  This also ensures that the object is constructed in a consistent way.

cr0wst:
This PR is a bit older so these comment may not be as relevant:

1. It looks like `baseUri` is becoming a common property between each of the endpoint clients.  Since each client doesn't currently have a parent class, this could be functionality that an abstract class provides through inheritance.

2. I also think it's interesting that the uri passed from the `NexmoClient` to the individual clients and then onto the endpoint.  It's stored on the `NexmoClient`, the individual clients, and eventually stored on the endpoints as part of `uri`.  It seems like there are a lot of places for this URI to be referenced.  Providing statefulness to so many locations doesn't seem completely necessary.  If I were to call `NexmoClient.setBaseUri` what is the gain from storing it on the client, or the individual clients, if it's just going to propagate and override the uri on the individual endpoints?

I also see this as a property of each endpoint, which would make it a good candidate (in my opinion) for something that is in `AbstractMethod` which should maybe just be called `Endpoint` (although I see that sometimes it's extended by a `*Method` and I haven't figured out the difference yet.

ChrisGuzman:
I'm in favor of a `AbstractMethod` approach.

judy2k:
Hey @ChrisGuzman - this PR has decayed, and I think a better approach is to change the Endpoint/Method instances in-place, rather than re-creating when the value changes.  I almost have a solution in #149 

Do you mind if we close this? 😬

PR #105, opened 2017-08-22
Add AccountClient and code for getBalance.

judy2k:
## Contribution Checklist
* [x] Unit tests!


judy2k:
This resolves #106

src/main/java/com/nexmo/client/account/BalanceRequest.java
```diff
@@ -0,0 +1,25 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package com.nexmo.client.account;
+
+public class BalanceRequest {
```
ChrisGuzman: Why is this an empty class? If it's not needed can we replace this with null or void in BalanceEndpoint?

src/test/java/com/nexmo/client/account/AccountClientTest.java
```diff
@@ -0,0 +1,44 @@
...
+    @Before
+    public void setUp() {
+        client = new AccountClient(null);
+        client.balance = mock(BalanceEndpoint.class);
+    }
+
+    @Test
+    public void testGetBalance() throws Exception {
+        client.getBalance();
+        verify(client.balance).execute();
```
ChrisGuzman: Should we add some more tests in the future or do you think this covers what's needed?

judy2k: This is just the test for AccountClient - and AccountClient's job is just to pass on the request to BalanceEndpoint - which has its own tests for functionality, below.

ChrisGuzman: Alrighty

src/main/java/com/nexmo/client/account/BalanceRequest.java
ChrisGuzman: @judy2k any thoughts on this?

ChrisGuzman: @judy2k I made a PR against this PR with with my suggestion on replacing the need for this class with void  https://github.com/Nexmo/nexmo-java/pull/117

PR #113, opened 2017-09-01
Add the endpoint to purchase a number

judy2k:
_Describe your changes here_

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/src/main/java/com/nexmo/client/numbers/BuyNumberEndpoint.java
+++ b/src/main/java/com/nexmo/client/numbers/BuyNumberEndpoint.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2011-2017 Nexmo Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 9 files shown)


src/main/java/com/nexmo/client/numbers/BuyNumberEndpoint.java
```diff
@@ -0,0 +1,73 @@
...
+        request.addParams(requestBuilder);
+        return requestBuilder;
+    }
+
+    @Override
+    public BuyNumberResponse parseResponse(HttpResponse response) throws IOException, NexmoClientException {
+        int statusCode = response.getStatusLine().getStatusCode();
+        String json = EntityUtils.toString(response.getEntity());
+
+        if (statusCode >= 400 && statusCode < 500) {
```
ChrisGuzman: I don't think we do this logic to check the status code in other Endpoints, why here?

judy2k: At a higher level, I'm swallowing the response object, because it usually doesn't contain anything useful, so raising an Exception here is extra important (because it's the only clue you'll get that the call failed)

Exception handling is something that is pretty bad in this library right now. I have plans to tackle it in the AbstractMethod class. It'll be a backwards-incompatible change though, so I need to make sure I get it right first(ish) time.

judy2k:
These changes have been merged into #114

PR #116, opened 2017-09-14
Modify call extension

judy2k:
Add the ability to transfer calls, earmuff, mute, etc.

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/src/main/java/com/nexmo/client/voice/CallModifier.java
+++ b/src/main/java/com/nexmo/client/voice/CallModifier.java
@@ -30,14 +30,14 @@ public class CallModifier {
     private final String uuid;
     private final ModifyCallPayload modifyCallPayload;
 
-    CallModifier(String uuid, ModifyCallPayload modifyCallPayload) {
+    public CallModifier(String uuid, ModifyCallPayload modifyCallPayload) {
         this.uuid = uuid;
         this.modifyCallPayload = modifyCallPayload;
     }
 
-    public CallModifier(String uuid, String action) {
+    public CallModifier(String uuid, ModifyCallAction action) {
         this.uuid = uuid;
-        this.modifyCallPayload = new ModifyCallPayload(ModifyCallPayload.Action.fromString(action));
+        this.modifyCallPayload = new ModifyCallPayload(action);
     }
 
     public static CallModifier transferCall(String uuid, String nccoUrl) {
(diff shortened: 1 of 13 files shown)


src/main/java/com/nexmo/client/voice/CallModifier.java
```diff
@@ -26,26 +26,30 @@
...
-    private ModifyCallPayload modifyCallPayload;
+    private final String uuid;
+    private final ModifyCallPayload modifyCallPayload;
+
+    CallModifier(String uuid, ModifyCallPayload modifyCallPayload) {
+        this.uuid = uuid;
+        this.modifyCallPayload = modifyCallPayload;
+    }
 
     public CallModifier(String uuid, String action) {
```
ChrisGuzman: Why not make the `action` an enum as well instead of a string?

src/main/java/com/nexmo/client/voice/CallModifier.java
```diff
@@ -26,26 +26,30 @@
...
 
     public String getUuid() {
         return uuid;
     }
 
     public String getAction() {
-        return modifyCallPayload.getAction();
-    }
-
-    public void setAction(String action) {
```
ChrisGuzman: Why remove the setter?

src/main/java/com/nexmo/client/voice/CallModifier.java
judy2k: Backwards incompatible change - I thought about it.

judy2k: (Backwards incompatible change to a _public interface_)

src/main/java/com/nexmo/client/voice/CallModifier.java
judy2k: I couldn't see any good reason to have it there. Given it's a backwards incompatible change, I should probably have left it, but this PR introduces a backwards incompatible change anyway, so we're going up to version 3 whether we like it or not :)

src/main/java/com/nexmo/client/voice/CallModifier.java
judy2k: Actually, given this PR introduces a backwards incompatible change, maybe we should just bite the bullet and do it now.

judy2k:
Approved offline by @ChrisGuzman

PR #117, opened 2017-09-19
Use void in BalanceEndpoint

ChrisGuzman:


judy2k:
Nice! Didn't know you could do this.

PR #123, opened 2017-10-12
Http error responses

judy2k:
Http Error status codes should raise HttpResponseException. Now they do.

An HTTP 400 (bad request) and 420 (Enhance your calm) are used to indicate bad requests, so in some cases those are specifically found and thrown as NexmoBadRequestException. In other error cases, and other endpoints which have never special-cased 420, an HttpResponseException is now thrown.

## Contribution Checklist
* [x] Unit tests!


src/test/java/com/nexmo/client/verify/VerifyClientTest.java
```diff
@@ -154,7 +154,7 @@ public void testCheckError() throws Exception {
     @Test
     public void testCheckInvalidResponse() throws Exception {
         wrapper.setHttpClient(this.stubHttpClient(
-                500,
```
ChrisGuzman: Why change from 500 to 200?

src/test/java/com/nexmo/client/verify/VerifyClientTest.java
```diff
@@ -390,7 +390,7 @@ public void testSearchDodgyDates() throws Exception {
 
     @Test
     public void testSearchHttpError() throws Exception {
-        wrapper.setHttpClient(this.stubHttpClient(500, "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
```
ChrisGuzman: Why change from 500 to 200?

judy2k: The 500 was a mistake in the original test. It's meant to be testing a failed response parse, which is why it throws/catches NexmoResponseParseException

PR #139, opened 2018-01-23
Convert SMS Endpoint from XML to JSON

cr0wst:
Updated the SMS endpoint to use the JSON endpoint instead of the XML endpoint.  This introduces a new response class `SmsSubmissionResponse` and `SmsSubmissionResponseMessage` to mirror how some of the other JSON endpoints operate.

`SmsSubmissionResponse` contains a `Collection<SmsSubmissionResponseMessage>` which replaces the `SmsSubmissionResult[]`.  This breaks backwards compatibility but it is much more flexible for the end user.

## Contribution Checklist
* [x] Unit tests!


ChrisGuzman:
Thanks so much! Looks great. Let's see what Mark thinks after he gets back from PTO

judy2k:
This is really nice. It _is_ a backwards-incompatible change though, so I may hold off on releasing for the time being, so we can bundle it together with some other breaking changes.

PR #148, opened 2018-02-25
Add level property to TalkNcco

### LINKED ISSUE #146: Add the "level" attribute to the talk objects
Reading the documentation it seems that the level attribute (see below) is missing in the TalkNcco object and in the VoiceClient.startTalk methods.

> The volume level that the speech is played. This can be any value between -1 to 1 with 0 being the default.

cr0wst:
Added the level property to `TalkNcco` to match the [NCCO Reference](https://developer.nexmo.com/api/voice/ncco)

I also split the tests out in `TalkNccoTest` because they were using the same process (Jackson) to serialize the object into JSON and then unserialize it back into the object and testing the results.  I felt like using Jackson to create the JSON wasn't providing adequate testing in regards to the expected JSON format.

## Contribution Checklist
* [x] Unit tests!

Resolves #146 ? I didn't see this attribute in the [Voice API Reference](https://developer.nexmo.com/api/voice) when starting a call.


src/main/java/com/nexmo/client/voice/ncco/TalkNcco.java
```diff
@@ -63,6 +66,14 @@ public void setLoop(Integer loop) {
         this.loop = loop;
     }
 
+    public void setLevel(BigDecimal level) {
```
judy2k: Do we need the level of accuracy `BigDecimal` provides? That type is really designed for handling currency or something where precision is super-important. I would have expected to see a `Float` type here.

src/test/java/com/nexmo/client/voice/ncco/TalkNccoTest.java
```diff
@@ -21,35 +21,65 @@
...
+        ncco.setLevel(new BigDecimal("-0.34"));
+        assertEquals(json, ncco.toJson());
+    }
+
+    @Test
+    public void testJsonHasVoiceNameWhenProvided() {
+        String json = "{\"text\":\"Talk to me\",\"voiceName\":\"Kimberly\",\"action\":\"talk\"}";
+        TalkNcco ncco = new TalkNcco("Talk to me");
+        ncco.setVoiceName("Kimberly");
+        assertEquals(json, ncco.toJson());
```
judy2k: Nice test refactoring! Thanks!

src/main/java/com/nexmo/client/voice/ncco/TalkNcco.java
cr0wst: That's a fair point.  I wasn't sure what level of precision would be necessary here and was erring on the side of more precise especially when working with the -1 to 1 range.  I think a `Float` is totally sufficient for this.

PR #152, opened 2018-03-04
StreamNcco serialization fix

ruXlab:
As per [documentation](https://developer.nexmo.com/api/voice/ncco#stream) `streamUrl` field of NCCO has to be an array with 1 url. At the moment value of `streamUrl` is serialised as a string which causes error.

Sample from unit test doesn't seem to be working in Voice Playground as well:  
screenshot from 2018-03-04 01-48-16

Given solution will not break compatibility with existing code.

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Modify call now works with the API again, returning `null` (because the API now returns 204 No Content)
 - VerifyRequest now supports all supported parameters.
+- `StreamNcco` serializes/deserializes `streamUrl` field correctly as per documentation
 
 ## [3.2.0] - 2017-11-30
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -7,3 +7,4 @@ documentation, or tests are eligible to be added to this list.
 - Chris Guzman ([@ChrisGuzman](https://github.com/ChrisGuzman))
 - Tim Lytle ([@tjlitle](https://github.com/tjlitle))
 - [@thepeachbeetle](https://github.com/thepeachbeetle)
+- Ruslan Zaharov ([@ruXlab](https://github.com/ruXlab))
--- a/src/main/java/com/nexmo/client/voice/ncco/StreamNcco.java
+++ b/src/main/java/com/nexmo/client/voice/ncco/StreamNcco.java
@@ -41,12 +41,12 @@ public class StreamNcco implements Ncco {
     }
 
     @JsonSetter("streamUrl")
-    void setStreamUrlArray(String[] streamUrl) {
+    private void setStreamUrlArray(String[] streamUrl) {
         this.streamUrl = streamUrl;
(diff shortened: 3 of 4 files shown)


cr0wst:
What are your thoughts on leaving it as a `String` on the `StreamNcco` class, but having Jackson use a different getter to convert it to a single-value array?

ruXlab:
Hey, @cr0wst I see what you mean. 
Is there specific reason for that? Generally speaking these internals are hidden from end users so it won't make big difference :)

cr0wst:
Hi @ruXlab, I guess I don't have a *strong* reason.  It's very true that the underlying type is being hidden and it won't make a huge difference.

I see the Jackson mapping process as a presentation-level process.  I guess my thought was based on the fact that it seems like we are manipulating the underlying data structure, instead of modifying the presentation process.  I don't see any other gains from storing the value as a fixed array of length 1.

Again, I don't have a strong feeling either way; I was more curious what your thoughts were. :)

ruXlab:
@cr0wst yeah, I don't have problems with changing on representation level either :)   
I'd like to see internal representation reflecting actual json format while exposing easy to use properties. Again both solutions seem equal in general :100: 

Btw what do you think about builders for Ncco? I see fluent interface would work here well.

cr0wst:
Are you talking about the various `Ncco` implementations, and not just `StreamNcco`?  Perhaps building objects that adhere to the `Ncco` interface, without having to implement each action as its own class?

I think, if that's the case, there's certainly a lot of room for clean-up, although I'm not 100% sure how Jackson would play with them, it may require a custom serializer. 👍 

I'm usually in the more *traditional* Java Bean camp, using builders when it's important to maintain immutability, or maybe when there are a lot of required parameters.  Is there any sort of advantage you see to them that I might be missing?

That being said, I'm not against it and working with PHP/Laravel and Java's streams has certainly started to change my mind. :)  Maybe an `NccoBuilder` that would allow you to chain specific actions? Hmm.

judy2k:
Hello! I _do_ have reasonably strong opinions on it :)

The internal representation should be as "correct" as possible, by which I mean it shouldn't be restricted to the set of flawed types that JSON makes available. JSON should be converted to whatever is the best representation available. This means ISO-8601 strings should be converted to Date objects, and booleans should be converted to `true` and `false`, for example.

ruXlab:
@cr0wst regarding builders. They can be used to construct object, not as DTO

For example:
```java
StreamNcco ncco = new StreamNcco(url);
ncco.setLevel(0.5);
ncco.setBargeIn(true);
// use ncco here
```

can be converted to 
```java
StreamNcco ncco = new StreamNcco.Builder(url).level(0.5).bargeIn(true).build();
```

Alternatively, `StreamNcco`'s setters can return `StreamNcco` object itself, so it's possible to chain setters:
```java
StreamNcco ncco = new StreamNcco(url).setLevel(0.5).setBargeIn(true)
```
In a way it's similar to kotlin constructors where it's possible to provide labelled parameters

Issue #147, opened 2018-02-24
add the HUMAN enum to CallStatus


blaluc:

cr0wst:
This was fixed in #191.

Prior to the 3.8.0 release we didn't have objects to handle the deserialization of the json payload sent to the `event_url` in various NCCO actions. Our documentation implies that the only way for a `human` status to occur is through the request sent to the `event_url`. Please feel free to re-open and correct me if I'm wrong.

The `com.nexmo.client.voice.CallStatus` enum was used with `CallEvent` created by `VoiceClient#createCall` and `CallInfo` created by `VoiceClient#listCalls`. It was never intended to be used for this purpose, but I understand where you're coming from because the name is used in two different parts of the API.

In #191 I created a separate package `com.nexmo.client.incoming` which has classes to achieve what, I think, you're trying to achieve.

PR #193, opened 2018-09-25
Bugfix/147 add human to callstatus

christiffer:
Bugfix for [#147](https://github.com/Nexmo/nexmo-java/issues/147) : Added HUMAN to CallStatus for voice machine detection.


cr0wst:
Hi @christiffer,

Thanks for your contribution. I have updated #147 to explain why this PR has been closed. To my knowledge, the original issue was trying to use the `CallStatus` enum in a way that was not intended.

Instead of adding `HUMAN` to this enum and use it in another way (which I think leads to confusion) I have added it to a separate one as part of the 3.8.0 release.

Issue #61, opened 2017-04-12
`AbstractMethod` is in the wrong package.


judy2k:
`AbstractMethod` is in `com.nexmo.client.voice.endpoints`, but is shared across all the APIs, so should be under `com.nexmo.client` or a common package.

ferozbaig96:
Can I work on this issue?

cr0wst:
Hi @arpitpandey0209, feel free to submit a PR for this issue. I would recommend starting with the `version-4` branch as this will most likely be a backwards incompatible change and would need to be released as part of the next major version.

Soundug:
Hi @cr0wst I would like to work on this issue. Is it up for grabs?

cr0wst:
Resolved in #225

PR #213, opened 2018-10-24
Move AbstractMethod to the client package

Soundug:
Updated packages of AbstractMethod and AbstractMethodTest

## Contribution Checklist
* [ ] Unit tests!
* [ ] My name is in [CONTRIBUTORS.md](CONTRIBUTORS.md)


Soundug:
Fixed #61

cr0wst:
There are a few things that would need to be done in addition to just moving the class:

1. You will need to apply this change to the `version-4` branch as it breaks backwards compatibility and will not be able to be released in 3.x.
2. You will also need to update references to the class because they will be using the old package.

I would suggest checking out a branch off of `version-4` and performing the refactor.

PR #233, opened 2019-03-12
Create Call with NCCO

### LINKED ISSUE #230: VoiceClient createCall API does not accept NCCO
When creating a call with `VoiceClient.createCall()`, the `Call` object does not accept a `NCCO`. But according to the documents, we can send a `NCCO` instead of `answered_url`. How can I resolve this issue?

Nexmo doc ref: https://developer.nexmo.com/api/voice

cr0wst:
You can now call the `createCall` method using a `Call` object containing an `Ncco`.

## Contribution Checklist
* [x] Unit tests!

Resolves #230

### DIFF
--- a/src/main/java/com/nexmo/client/voice/Call.java
+++ b/src/main/java/com/nexmo/client/voice/Call.java
@@ -58,9 +58,7 @@ public class Call {
     }
 
     public Call(Endpoint to, Endpoint from, String answerUrl) {
-        this.to = new Endpoint[]{to};
-        this.from = from;
-        this.answerUrl = answerUrl;
+        this(new Endpoint[]{to}, from, answerUrl);
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/nexmo/client/voice/Call.java
```diff
@@ -84,7 +103,7 @@ public void setFrom(Endpoint from) {
 
     @JsonProperty("answer_url")
     public String[] getAnswerUrl() {
-        return new String[]{answerUrl};
+        return (answerUrl == null) ? null : new String[]{answerUrl};
```
cr0wst: This should probably be consistent with `getAnswerMethod`.

```suggestion
        return (answerUrl == null) ? new String[]{answerUrl} : null;
```

Issue #237, opened 2019-04-02
Message product's spelling is wrong for Redact API 


sharif-malik:
After talking to support team, I got to know that  **message** should **messages** in the RedactRequest.Product enum.

sharif-malik:
https://github.com/Nexmo/nexmo-java/blob/master/src/main/java/com/nexmo/client/redact/RedactRequest.java#L90

sharif-malik:
Here is the documentation from NEXMO(https://developer.nexmo.com/api/redact), which is expecting **messages**

sharif-malik:
#238 Here is the PR to fix this issue

sharif-malik:
@cr0wst Please review this Issue and followed by PR #238

sharif-malik:
Backward compatibility, actually failing my tests when I do integration tests.

sharif-malik:
@cr0wst , it should be `message` instead of `messages`

sharif-malik:
Shall I create another PR for the same. Because currently my tests is picking from text so it always pick `MESSAGE` instead of `MESSAGES`

sharif-malik:
@cr0wst Here is the another PR to fix the value of deprecated product name value

sharif-malik:
Because it fails everytime to pick right product enum

cr0wst:
We only use this object for serialization, it's not intended to be used for deserialization at this time.

sharif-malik:
Yeah, I understand your point. Currently having both enum's value as same is not resilient, I am not able to do integration tests in my project because of deserialisation.

This #240  PR, it will work for all. Isn't it?

cr0wst:
Added explanation in #240.

PR #240, opened 2019-04-03
Fix deprecated product name for message

sharif-malik:
Fix product name for deprecated product name. 

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,9 +1,9 @@
 [bumpversion]
 commit = True
 tag = True
-current_version = 4.2.1
+current_version = 4.2.2
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
-serialize = 
+serialize =
 	{major}.{minor}.{patch}
 
 [bumpversion:file:build.gradle]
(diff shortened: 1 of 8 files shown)


sharif-malik:
@cr0wst Could you please review this PR?

cr0wst:
Hi @sharif-malik,

As I explained in #237, the purpose of `RedactRequest` and it's associated enum is to create JSON for the Nexmo API to consume. It is not intended to work in reverse, so the ability for Jackson to use it in deserialization is not a significant factor in deciding to make this change.

I understand that you are trying to use this class in your integration testing, and you are correct that the robustness is questionable when working in both serialization and deserialization. However, our only concern with this class is that it can create the correct JSON payload for the API. This concern is why our testing only exercises its ability to create JSON, and not also its ability to deserialize the JSON back into the object format. I appreciate the earlier work you did to identify a bug in which it was incorrectly doing this.

Using this class in a way that it is not intended, especially in the context of your integration tests, isn't something that we want to encourage as it will require a lot more resources to monitor and fix.

See #147 in which another user was attempting to use `CallStatus` and `CallInfo` in an unintended way.

sharif-malik:
Thank you for your response. I understand your concern regarding intent of from java to json. 
I agree with your comment upto some extent but I don't find any harm to have both ways(serialization and deserialization). 

FYI, I don't want to finger point but the deserialization was broke the moment wrong deprecated value was introduced. Earlier it was `message` and now its `messages`. If you are adding deprecated it should remain to the old value. Isn't it? 
Addition to it, so far, all the models are both way compatible. Isn't it good to remove typo mistake and make the code better in order to maintain in future and can help the client to build more quality of code rather half way support.

PR #250, opened 2019-05-31
Add support for Applications v2

cr0wst:
Change Applications to use v2. This breaks backwards compatibility.

## Contribution Checklist
* [x] Unit tests!
* [x] Updated [CHANGELOG.md](CHANGELOG.md)
* [x] My name is in [CONTRIBUTORS.md](CONTRIBUTORS.md)

PR #272, opened 2019-11-04
[DO NOT MERGE] Conversations updates

slorello89:
_Describe your changes here_

## Contribution Checklist
* [ ] Unit tests!
* [ ] My name is in [CONTRIBUTORS.md](CONTRIBUTORS.md)

### DIFF
--- a/src/main/java/com/nexmo/client/NexmoClient.java
+++ b/src/main/java/com/nexmo/client/NexmoClient.java
@@ -25,6 +25,7 @@ package com.nexmo.client;
 import com.nexmo.client.account.AccountClient;
 import com.nexmo.client.application.ApplicationClient;
 import com.nexmo.client.auth.*;
+import com.nexmo.client.conversation.ConversationClient;
 import com.nexmo.client.conversion.ConversionClient;
 import com.nexmo.client.insight.InsightClient;
 import com.nexmo.client.numbers.NumbersClient;
@@ -60,6 +61,7 @@ public class NexmoClient {
     private SnsClient sns;
     private ConversionClient conversion;
     private RedactClient redact;
+    private ConversationClient conversations;
     private HttpWrapper httpWrapper;
 
     private NexmoClient(Builder builder) {
@@ -76,6 +78,7 @@ public class NexmoClient {
         this.sns = new SnsClient(this.httpWrapper);
         this.conversion = new ConversionClient(this.httpWrapper);
         this.redact = new RedactClient(this.httpWrapper);
+        this.conversations = new ConversationClient(this.httpWrapper);
     }
 
     public AccountClient getAccountClient() {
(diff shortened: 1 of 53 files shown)


cr0wst:
Hi! I've been on a bit of a GitHub hiatus, but I noticed this PR come through on my notifications. It seems like it's a work in progress, and I wanted to leave you with a few things you might find useful.

First, all of the HATEOAS stuff can be included by changing up your serializer. Instead of using the `ObjectMapper` from Jackson, you can use the `HALLink` and `@EmbeddedResource` annotation provided by the included `'io.openapitools.jackson.dataformat:jackson-dataformat-hal:1.0.4'` library along with the `HALMapper`. This will generate a lot of the embeddable relationships without you having to worry about crafting the structure yourself. If the structure is a bit more complicated, it might take some additional configuration options.

Pull request #250 shows how it's used in Application V2. I haven't seen any specifications for the Conversation service, so it might not be similar enough for a reference.

https://github.com/Nexmo/nexmo-java/blob/6a0fa2aa22cd27bbb17359d6ecf168cb2e47ae38/src/main/java/com/nexmo/client/application/ApplicationList.java#L39-L40

https://github.com/Nexmo/nexmo-java/blob/6a0fa2aa22cd27bbb17359d6ecf168cb2e47ae38/src/main/java/com/nexmo/client/application/ApplicationList.java#L47-L51

Secondly, since you're using `@JsonProperty`, you don't need to match the snake-case from the API on your member variables. I would also suggest that you don't use the primitive types and instead opt for the wrapper classes for your data structures as it allows you to hide them with Jackson's `JsonInclude.Include.NON_NULL`. That way, you can still allow for things like 0 without Jackson thinking that you don't want to include it in the resulting JSON.

Some of these items might also be available via the `PageList` class which is extendable:
https://github.com/Nexmo/nexmo-java/blob/6a0fa2aa22cd27bbb17359d6ecf168cb2e47ae38/src/main/java/com/nexmo/client/common/PageList.java#L28-L52

A third thing I wanted to mention was that in some places, you are an `Object` type. Namely in the `Event` class. The docs aren't very clear at the moment what exactly this is supposed to be. However, with Jackson, you can serialize `Map<String, ?>` to give you specific key-value pairs. If your goal is to map a JSON object, I think it's a reasonably safe way to go about doing it. This is how it was handled in the `NotificationAction` serializer

https://github.com/Nexmo/nexmo-java/blob/7427eff6d6baa5a5bd9197fa096bcf2564191cf5/src/main/java/com/nexmo/client/voice/ncco/NotifyAction.java#L39

It's a bit clunky during serialization, but users will have to cast in those cases anyway. I believe casting to a Map will at least allow them to iterate through.

I didn't mean to pry and was trying to offer a few, hopefully, helpful suggestions to make things a bit easier!

Have a beautiful day!

Issue #275, opened 2019-12-12
No option for selecting workflow 6 i.e. only SMS from VerifyRequest.Workflow


anmol5varma:
https://developer.nexmo.com/verify/guides/workflows-and-events
Workflow 6 is absent from `VerifyRequest.Workflow` Class. How do I implement that using nexmo java sdk?

lornajane:
Thanks for reaching out to us about this, we've just released a new version of the SDK that adds support for these workflows. Please can you try installing version 5.2.1 and let us know how you get on?

lornajane:
Closing, but re-open if anything more is needed here

PR #276, opened 2019-12-13
adding new workflow IDs, bumping version, updating changelog

slorello89:
_Describe your changes here_
Added Workflow IDs for SMS and TTS
## Contribution Checklist
* [x] Unit tests!


lornajane:
See also related question/issue #275

slorello89:
> Drop the `.DS_Store` file and it should be good

done

lornajane:
@dragonmantank did both workflows work as expected for you? id 6 is strange for me, it doesn't send an SMS

lornajane:
I'm struggling with this for two reasons:
 - Verify doesn't seem to work that well for me. I do sometimes have issues with it but my curl commands all seem to be going really well just now so I'm not certain where the problem is. If anyone else has this working fine, I'm happy to take that approval but as things stand I have not seen this operate correctly. Which might well be the API rather than this code!
 - The snippet should accept a `WORKFLOW_ID` which is an integer, having to look up what constant we chose instead doesn't feel like a great experience. Can we just make this an integer variable that gets passed in?

slorello89:
@lornajane - not sure why it's acting quirkily for you from the APIs perspective - the 6 & 7 worked well for me. In regards to the constants - this specifically only accepts an enum which is descriptive of what the workflow does. You can cast an ordinal to that enum but if you're programming agasint this you're going to want to use the enum

dragonmantank:
@lornajane Just checked against the code snippets, both SMS and TTS worked fine.

I have had issues where if I've sent too many Verify requests without canceling/verifying them it slows down and I get an angry e-mail from Nexmo.

dragonmantank:
@lornajane @slorello89 Could we update the code snippets to show the casting usage so that we're compliant with the code snippet specs? Then at least we have it documented how we expect people to use the raw Worfklow ID coming from the API reference.

lornajane:
That works for me, please merge and make sure we reply to the slack thread where this was asked - there's an open customer ticket waiting on this. Thanks :)

PR #285, opened 2020-06-08
Add language, gender, SSML support to VoiceName enum

sharmasha2nk:
Add language, gender, SSML support to VoiceName enum and deprecated Chipmunk, Eric, Jennifer, and Agnieszka


yallen011:
It was determined that the case for things like randomizing voices by languages and such were more of an edge case and would require much more overhead to change at this point

PR #297, opened 2020-08-13
Bugfix/overridable base urls

### LINKED ISSUE #294: ModifyCallMethod and SendDtmfMethod not using HttpConfig for base URLs
Not able to use alternate base URLs for Modify call and send DTMF methods. Need to update the base URLs to handle [this known issue](https://help.nexmo.com/hc/en-us/articles/115015969628-Why-do-I-get-a-404-when-trying-to-change-an-active-conversation-?source=search).

DEFAULT_URI are hardcoded in both these classes.

yallen011:
Fixes #294 

- Make base uri overridable
- Add test for stop text to speech request during a call
- Add test for stop stream audio request during a call

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,14 +7,14 @@ buildscript {
     }
 }
 
-//apply plugin: 'com.github.johnrengelman.shadow'
+apply plugin: 'com.github.johnrengelman.shadow'
 
 apply plugin: 'java'
 apply plugin: "jacoco"
 apply plugin: 'maven'
 apply plugin: 'signing'
 apply plugin: 'eclipse'
-// apply plugin: 'war'
+
 
 group = "com.nexmo"
 archivesBaseName = "client"
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
     }
 }
 
-apply plugin: 'com.github.johnrengelman.shadow'
+//apply plugin: 'com.github.johnrengelman.shadow'
 
 apply plugin: 'java'
 apply plugin: "jacoco"
--- a/src/main/java/com/nexmo/client/voice/ModifyCallMethod.java
+++ b/src/main/java/com/nexmo/client/voice/ModifyCallMethod.java
@@ -38,9 +38,8 @@ import java.io.UnsupportedEncodingException;
 class ModifyCallMethod extends AbstractMethod<CallModifier, ModifyCallResponse> {
     private static final Log LOG = LogFactory.getLog(ModifyCallMethod.class);
 
-    private static final String DEFAULT_URI = "https://api.nexmo.com/v1/calls/";
+    private static final String PATH = "/calls/";
     private static final Class[] ALLOWED_AUTH_METHODS = new Class[]{JWTAuthMethod.class};
-    private String uri = DEFAULT_URI;
 
     ModifyCallMethod(HttpWrapper httpWrapper) {
         super(httpWrapper);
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
     }
 }
 
-apply plugin: 'com.github.johnrengelman.shadow'
+//apply plugin: 'com.github.johnrengelman.shadow'
```
dragonmantank, line 10: Was this plugin causing problems, or just no longer needed? Instead of commenting it out, we can just delete it entirely. 

Do we not use it to create the JAR files?

src/main/java/com/nexmo/client/voice/ModifyCallMethod.java
```diff
@@ -53,9 +52,8 @@
 
     @Override
     public RequestBuilder makeRequest(CallModifier request) throws UnsupportedEncodingException {
-        String uri = this.uri + request.getUuid();
         return RequestBuilder
-                .put(uri)
+                .put(httpWrapper.getHttpConfig().getVersionedApiBaseUri("v1") + PATH + request.getUuid())
```
dragonmantank, line 56: Is this something we should make sure we are doing for all the other products as well, or did we just miss this for a few endpoints on Voice?

yallen011: This was missed in only a few of the endpoints. all of the rest do this except for the 3 files that are in this PR.

build.gradle
yallen011: I currently causes problems. I believe we create jars without it but I will check again and if we do, I will remove it completely.

yallen011: fixed, it only causes a problem when I am using it to test the code snippets locally for some reason.

dragonmantank:
Unit tests do not pass on OpenJDK 1.8

```
> Task :test 

com.nexmo.client.voice.StopTalkMethodTest > makeRequestTest FAILED
    java.lang.AssertionError: 
        at org.junit.Assert.fail(Assert.java:74)
        at org.junit.Assert.assertTrue(Assert.java:37)
        at org.junit.Assert.assertTrue(Assert.java:46)
        at com.nexmo.client.voice.StopTalkMethodTest.makeRequestTest(StopTalkMethodTest.java:44)

644 tests completed, 1 failed, 1 skipped
```

src/test/java/com/nexmo/client/voice/StopStreamMethodTest.java
```diff
@@ -23,34 +23,88 @@
...
+    String httpMethod;
 
     @Before
     public void setUp() throws Exception {
         this.method = new StopStreamMethod(new HttpWrapper());
+        httpMethod = "DELETE";
+    }
+
+    @Test
+    public void makeRequestTest() throws UnsupportedEncodingException {
```
dragonmantank, line 55: Unit tests do not pass on OpenJDK 1.8

```
> Task :test 

com.nexmo.client.voice.StopTalkMethodTest > makeRequestTest FAILED
    java.lang.AssertionError: 
        at org.junit.Assert.fail(Assert.java:74)
        at org.junit.Assert.assertTrue(Assert.java:37)
        at org.junit.Assert.assertTrue(Assert.java:46)
        at com.nexmo.client.voice.StopTalkMethodTest.makeRequestTest(StopTalkMethodTest.java:44)

644 tests completed, 1 failed, 1 skipped
```

yallen011: I think you put this comment on the wrong file. but I do see 2 failing now.

yallen011: fixed failing test.

PR #302, opened 2020-09-15
Made CI changes

yallen011:
- Added github actions
- Updated badges
- Removed Travis CI


slorello89:
I don't see a GitHub action kicking off?

yallen011:
could that be because it is not in master yet?

.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{matrix.java}}
+
+      - name: Run Codecov
```
dragonmantank, line 20: While Travis CI didn't seem to do it either, can we add a step to run the unit tests and make sure that the library compiles?

README.md
```diff
@@ -1,10 +1,10 @@
 # Vonage Server SDK for Java
 
-[![Maven Release](https://maven-badges.herokuapp.com/maven-central/com.nexmo/client/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.nexmo/client)
-[![Build Status](https://travis-ci.org/Nexmo/nexmo-java.svg?branch=version-2)](https://travis-ci.org/Nexmo/nexmo-java)
-[![codecov](https://codecov.io/gh/Nexmo/nexmo-java/branch/version-2/graph/badge.svg)](https://codecov.io/gh/Nexmo/nexmo-java)
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/0049e762c00b4ce999f45492795ba50a)](https://www.codacy.com/app/cr0wst/nexmo-java?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=Nexmo/nexmo-java&amp;utm_campaign=Badge_Grade)
-
+[![Maven Release](https://maven-badges.herokuapp.com/maven-central/com.vonage/client/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.vonage/client)
+[![Build Status](https://github.com/vonage/vonage-java-sdk/workflows/build/badge.svg)](https://github.com/Vonage/vonage-java-sdk/actions?query=workflow%3Abuild)
```
dragonmantank, line 4: The path for the badge and link need to use the `name` value used in the YAML file, not the filename itself.

.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
+name: Java Core
```
dragonmantank, line 1: Can we have this more in-line with what this file is doing? Calling it "Build" or "CI" or something more descriptive than "Java Core" would be useful as we add more workflows down the road

.github/workflows/build.yml
```diff
@@ -0,0 +1,21 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
dragonmantank, line 6: Can we expand this to Windows and MacOS?

codecov.yml
```diff
@@ -1,20 +1,12 @@
 codecov:
-  branch: version-2
+  branch: master
   bot: judy2k
```
dragonmantank, line 3: Can we change this value or remove it?

dragonmantank (changes requested):

README.md
yallen011: fixed

.github/workflows/build.yml
yallen011: fixed

.github/workflows/build.yml
yallen011: added

codecov.yml
yallen011: Just for clarity, why do you want it to be removed?

dragonmantank: He no longer works here, so we don't need his naming showing up in automated messages or notifications.

yallen011: I don't know how to create a different bot. Removing that bot will cause a different bot to be used and I don't know what that bot will do or what will be different. I think it will use the team bot tho, which I don't think is set.

slorello89:
@yallen011 - It looks like the windows machine is not replacing the `-----BEGIN PRIVATE KEY-----` correctly from the keys and a `-` is bleeding into the private key it's trying to parse (hence the 2d illegal character - which it's reading as `2d`).

My guess is that this is an issue in the [JWT library](https://github.com/Nexmo/nexmo-jwt-jdk/blob/master/src/main/kotlin/com/nexmo/jwt/KeyConverter.kt#L44) - which is assuming the line ending type of the first line.

see: https://www.prostdev.com/post/understanding-the-illegal-base64-character-error-java-groovy-and-mule-4-dw-2-0

dragonmantank (approved):

Issue #229, opened 2019-01-22
Support signature generation using algorithms other than MD5-Hash


judy2k:
It looks like RequestSigning only supports MD5-Hash. I think this is an oversight, and the HMAC signing methods should be supported as well.

Raj-Datta-Manohar:
Hi, I would like to work on this issue. I would like to know how to determine which signing method is to be used(md5 or sha1 or sha256).

yallen011:
Sure, have at it.

PR #341, opened 2020-10-06
Create SHA256 Hashing option

sirNikolai:
Addresses issue #229 
Make a more one way approach to retrieving hashes and add another builtin java security hashing algorithm SHA-256.
Also in creating the abstraction, simplifies the way a bit when needing to creating different hash types. While the current 2 look very similar (only change is basically the Message Digest) this adds capability for create more with varying components (e.g non-java native hashing functions).

### Potentially Breaking Changes
- MD5Util was moved and renamed (can put back the file and link it to new internal API)
- SignatureAuthMethod had an argument added to constructor so now is also breaking (can be overloaded)

## Contribution Checklist
* [x] Unit tests!


src/main/java/com/vonage/client/auth/hashutils/MD5Util.java
```diff
@@ -25,17 +25,17 @@
  *
  * @author Paul Cook
  */
-public class MD5Util {
+class MD5Util extends Hasher {
```
sirNikolai, line 28: My only qualm with this is hoping that it did not break any API agreements that the SDK has. Otherwise all other methods are refactored to work as intended with this as an added feature without breaking any contracts.

JamiesWhiteShirt:
I can't seem to figure out how you would configure a `VonageClient` to use a different signature with these changes. The changes seem to be mostly internal. Should perhaps a `HashUtil.HashType` be configurable in the `VonageClient.Builder`?

sirNikolai:
I see what you mean. I'll change it up in the regard.

src/main/java/com/vonage/client/VonageClient.java
```diff
@@ -144,6 +145,7 @@ public static Builder builder() {
         private String apiSecret;
         private String signatureSecret;
         private byte[] privateKeyContents;
+        private HashUtil.HashType hashType = HashUtil.HashType.MD5;
```
sirNikolai, line 148: main reason for the default is to prevent existing API breakages.

src/main/java/com/vonage/client/auth/SignatureAuthMethod.java
```diff
@@ -25,17 +26,19 @@
 
     private String apiKey;
     private String secret;
+    private HashUtil.HashType hashType;
 
-    public SignatureAuthMethod(String apiKey, String secret) {
+    public SignatureAuthMethod(String apiKey, String secret, HashUtil.HashType hashType) {
```
sirNikolai, line 31: potentially needs an overload but will leave to your judgement. Otherwise this could maybe be a breaking change.

sirNikolai:
This will require another refactor though to get HMAC SHA256 added in for example since right now the secret key is just appended to the end of the string when for HMAC it should be passed in separately. Can probably do it as part of this PR but again not sure what the bounds of this is.

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
```diff
@@ -0,0 +1,58 @@
...
+     * @throws NoSuchAlgorithmException if the algorithm is not available.
+     * @throws UnsupportedEncodingException if the specified encoding is unavailable.
+     */
+    public String calculate(String input, String secretKey, String encoding, HashType hashType) throws NoSuchAlgorithmException, UnsupportedEncodingException, InvalidKeyException {
+        return hashTypes.get(hashType).calculate(input, secretKey, encoding);
+    }
+
+    public enum HashType {
+        MD5,
+        SHA256,
```
slorello89, line 55: We don't support a straight SHA256 hash - we only support the [HMAC](https://developer.nexmo.com/concepts/guides/signing-messages).

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        HMAC_SHA256
```
slorello89, line 56: be really cool if we could get `HMAC_SHA512` as well (we also support SHA1 but I think it's fine not to include it)

src/main/java/com/vonage/client/auth/hashutils/MD5Hasher.java
```diff
@@ -49,20 +33,12 @@ public static String calculateMd5(String input) throws NoSuchAlgorithmException
      * @throws NoSuchAlgorithmException if the MD5 algorithm is not available.
      * @throws UnsupportedEncodingException if the specified encoding is unavailable.
      */
-    public static String calculateMd5(String input, String encoding) throws NoSuchAlgorithmException, UnsupportedEncodingException {
```
slorello89, line 52: `calculateMd5` should be left in - even if it's just being routed through calculate - removing it is a breaking change.

slorello89 (changes requested):

src/main/java/com/vonage/client/auth/RequestSigning.java
```diff
@@ -62,11 +64,47 @@ public static void constructSignatureForRequestParameters(List<NameValuePair> pa
...
+     * Hashing strategy is MD5.
+     *
+     * @param params List of NameValuePair instances containing the query parameters for the request that is to be signed
+     * @param secretKey the pre-shared secret key held by the client
      * @param currentTimeSeconds the current time in seconds since 1970-01-01
      *
      */
-     protected static void constructSignatureForRequestParameters(
-            List<NameValuePair> params, String secretKey, long currentTimeSeconds) {
+     protected static void constructSignatureForRequestParameters(List<NameValuePair> params,
```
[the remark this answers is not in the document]
sirNikolai: Refer to previous comment

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    private Map<HashType, AbstractHasher> hashTypes;
+
+    private HashUtil() {
+        hashTypes = new HashMap<>();
+        hashTypes.put(HashType.MD5, new MD5Hasher());
+        hashTypes.put(HashType.SHA256, new SHA256Hasher());
+        hashTypes.put(HashType.HMAC_SHA256, new HmacSha256Hasher());
+    }
+
+    public static HashUtil getInstance() {
```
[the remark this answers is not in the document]
sirNikolai: yeah, this can be a static helper class. In retrospect not sure why I decided on singleton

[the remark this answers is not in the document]
slorello89: 👍

src/main/java/com/vonage/client/auth/RequestSigning.java
```diff
@@ -122,15 +158,45 @@ public static boolean verifyRequestSignature(HttpServletRequest request, String
...
      *
-     * @param request The HttpServletRequest to be verified
-     * @param secretKey The pre-shared secret key used by the sender of the request to create the signature
+     * @param request The HttpServletRequest to be verified.
+     * @param secretKey The pre-shared secret key used by the sender of the request to create the signature.
+     *
+     * @return true if the signature is correct for this request and secret key.
+     */
+    public static boolean verifyRequestSignature(HttpServletRequest request, String secretKey, HashUtil.HashType hashType) {
+        return verifyRequestSignature(request, secretKey, System.currentTimeMillis(), hashType);
```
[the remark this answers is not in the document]
sirNikolai: Refer to first comment on optionals

src/main/java/com/vonage/client/auth/RequestSigning.java
```diff
@@ -62,11 +64,47 @@ public static void constructSignatureForRequestParameters(List<NameValuePair> pa
      *
      * @param params List of NameValuePair instances containing the query parameters for the request that is to be signed
      * @param secretKey the pre-shared secret key held by the client
+     * @param hashType The type of hash that is to be used in construction
+     */
+    public static void constructSignatureForRequestParameters(List<NameValuePair> params, String secretKey, HashUtil.HashType hashType) {
```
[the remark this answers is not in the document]
slorello89: TIL - cool (mostly a .NET dev so don't think I've ever noticed that about Java). Couple places where you've changed the signature rather than adding an override - in those spots definitely add an override so we don't induce a breaking change.

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
sirNikolai: Made that first before realising. I can either take it out entirely or keep it in and just remove the enum, making it unusable externally

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
sirNikolai: yeah can add that in lieu of just SHA 256

src/main/java/com/vonage/client/auth/RequestSigning.java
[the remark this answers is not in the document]
sirNikolai: documented that in the initial comment, was wondering if I should support that as legacy. I'll just re-add them back in and deprecate calculateMd5.

Also I do miss optionals and default values from .NET

src/main/java/com/vonage/client/auth/hashutils/HashUtil.java
slorello89: that would be awesome

src/main/java/com/vonage/client/auth/RequestSigning.java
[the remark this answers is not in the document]
slorello89: Yeah so these libraries are strict adherents to [semver](https://semver.org/) so breaking public methods is only really allowable between major versions. If your changes are making a method become obsolete feel free to tag it as such so people know which method to use but everything should be backward compatible so peoples code doesn't break 😃

src/main/java/com/vonage/client/auth/hashutils/AbstractHasher.java
```diff
@@ -16,7 +16,7 @@
     public String calculate(String input) throws NoSuchAlgorithmException {
         try {
             return calculate(input, "UTF-8");
-        } catch (UnsupportedEncodingException e) {
+        } catch (UnsupportedEncodingException | InvalidKeyException e) {
```
slorello89, line 19: Another spot where the exceptions should be advertised and thrown.

slorello89 (changes requested):
Hi @sirNikolai - Just a couple more notes - I think I gave you the wrong impression - we don't support a SHA-1 hash - we only supported it as an HMAC so I'm sorry 😭. I'm Going to try running these locally to confirm they work will let you know if I find anything else while I'm doing that. thanks!

sirNikolai:
@slorello89 cool cool. I'll wait for the local testing and then do the remaining changes once everything is done there.

src/main/java/com/vonage/client/auth/RequestSigning.java
```diff
@@ -174,25 +240,22 @@ protected static boolean verifyRequestSignature(HttpServletRequest request,
...
         }
 
-        log.info("SECURITY-KEY-VERIFICATION -- String [ " + str + " ] Signature [ " + md5 + " ] SUPPLIED SIGNATURE [ " + suppliedSignature + " ] ");
+        log.info("SECURITY-KEY-VERIFICATION -- String [ " + str + " ] Signature [ " + hashed + " ] SUPPLIED SIGNATURE [ " + suppliedSignature + " ] ");
 
         // verify that the supplied signature matches generated one
         // use MessageDigest.isEqual as an alternative to String.equals() to defend against timing based attacks
         try {
-            if (!MessageDigest.isEqual(md5.getBytes("UTF-8"), suppliedSignature.getBytes("UTF-8")))
+            if (!MessageDigest.isEqual(hashed.getBytes("UTF-8"), suppliedSignature.getBytes("UTF-8")))
```
slorello89, line 258: There's a bit of an issue here, for the HMACs they come in as uppercase, however, the hashed value generated is lower case - they're just integers so it ought to be case agnostic. This could be fixed as easily as sending both the hashed and supplied signature value to lowerCase before checking their equality. Unless you want to do something fancy 😄

slorello89:
@sirNikolai - I had one more comment after my testing, the verifyRequest doesn't work for HMACs as they come in from Vonage in upper case, and the calculate method produces the HMAC in lower case, they ought to be case agnostic but this always worked with the md5 hash because it just so happens to come in as a lower-case 🙃. If you want to add the two extra methods (HMAC SHA-1- and I forgot we also have HMAC-MD5) that would be much appreciated, otherwise I think what you've provided is easy enough to adapt to the other methods

sirNikolai:
@slorello89 wish I had the capacity to do something less cheaty to lowercase, buuuuuuut if it'll work then so be it

slorello89 (approved):
@sirNikolai - thank you so much for this - ran through send / receive on each of the different signing types and everything works perfectly - everything looks great from my perspective so I'm approving it. @yallen011 - I leave it to you to actually merge this - but this PR looks awesome.

thanks again Nik - Happy Hacktoberfest 👻 🎃 💀 !

sirNikolai:
Thanks! Hopefully this is trivial enough to extend if you enable more hashing types in the future.

PR #353, opened 2020-10-28
Add ASR

### LINKED ISSUE #288: Add ASR webhook for the Voice API
Add capability to handle Automatic Speech Recognition for voice inputs on a phone call. This will allow for using voice to specify an option from the menu given to you through the automated voice system.

yallen011:
- Add Automatic Speech Recognition feature to the SDK for the NCCO Input Action in the Voice API.

- Update SDK to use the new format for the DTMF settings in the Input Action NCCO.

Closes #288 

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [6.0.0]
+## Added
+- ASR (Automatic Speech Recognition)
+
 ## [5.6.0]
 ### Added
 - NotifyEvent structure for Notify Actions
--- a/src/main/java/com/vonage/client/incoming/InputEvent.java
+++ b/src/main/java/com/vonage/client/incoming/InputEvent.java
@@ -23,34 +23,66 @@ import com.vonage.client.VonageUnexpectedException;
 import java.io.IOException;
 import java.util.Date;
 
+
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class InputEvent {
     private String uuid;
     private String conversationUuid;
-    private boolean timedOut;
-    private String dtmf;
+    private DtmfResult dtmf;
     private Date timestamp;
+    private String to;
+    private String from;
+    private SpeechResults speech;
 
+    /**
+     * @return The unique identifier for this call
+     */
     public String getUuid() {
         return uuid;
     }
 
+    /**
+     * @return The unique identifier for this conversation
+     */
     @JsonProperty("conversation_uuid")
     public String getConversationUuid() {
         return conversationUuid;
     }
 
-    @JsonProperty("timed_out")
-    public boolean isTimedOut() {
-        return timedOut;
-    }
-
-    public String getDtmf() {
+    /**
+     * @return DTMF capturing retults.
+     */
+    public DtmfResult getDtmf() {
         return dtmf;
     }
 
+    /**
+     * @return Timestamp (ISO 8601 format)
+     */
     public Date getTimestamp() {
         return timestamp;
+    }
+
+    /**
+     * @return The number the call was made to
+     */
+    public String getTo(){
+        return to;
+    }
+
+    /**
+     * @return The number the call came from
+     */
+    public String getFrom(){
+        return from;
+    }
+
+    /**
+     * @return Speech recognition results
+     * @since 6.0.0
+     */
+    public SpeechResults getSpeech() {
+        return speech;
     }
 
     public static InputEvent fromJson(String json) {
--- a/src/main/java/com/vonage/client/voice/ncco/InputAction.java
+++ b/src/main/java/com/vonage/client/voice/ncco/InputAction.java
@@ -17,34 +17,35 @@ package com.vonage.client.voice.ncco;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
 import java.util.Arrays;
 import java.util.Collection;
 
 /**
- * An NCCO input action which allows for the collection of digits from a person.
+ * An NCCO input action which allows for the collection of digits and automatic speech recognition from a person.
  */
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class InputAction implements Action {
     private static final String ACTION = "input";
 
-    private Integer timeOut;
-    private Integer maxDigits;
-    private Boolean submitOnHash;
+    @JsonProperty(required = true)
+    private Collection<String> type;
+    private DtmfSettings dtmf;
     private Collection<String> eventUrl;
+    private SpeechSettings speech;
     private EventMethod eventMethod;
 
     /**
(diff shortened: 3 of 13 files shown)


src/main/java/com/vonage/client/incoming/InputEvent.java
```diff
@@ -17,46 +17,105 @@
...
+ *           the documentation.
+ * @since 5.6.0
+ */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class InputEvent {
+public class InputEvent<T> {
     private String uuid;
     private String conversationUuid;
-    private boolean timedOut;
-    private String dtmf;
```
slorello89, line 31: This is going to be a breaking change for some people it doesn't break the DTMF code snippet since they don't have to worry about the type in the string format. But anyone who's doing something like:

```java
String digits = inputtResult.getDtmf();
```

Is going to break. Unless they specify what the generic type is when they are creating the event, which they wouldn't have had to do before. 

You can keep the dtmf field as generic, but I think it might be wiser to have the getter be a string, and perhaps have other fields for timed out.

What are your thoughts @dragonmantank?

slorello89 (changes requested):

src/main/java/com/vonage/client/incoming/InputEvent.java
yallen011: ok, there was an assumption that if you are doing ASR you are going to use the generic but if you are doing DTMF the old way you wouldn't have to use the generic because dtmf would come back as a string which wouldn't through an error when you called `getDTMF()`. But if you are going to adapt the new way of send DTMF then you would use the generic. That was the assumption. I could add more verbiage to the javadoc to help with that but these are the assumptions made and this is the best trade-off I could find. This is not going to remain permanent since the plan is to remove the deprecated fields and the generic to just use DtmfResult as the datatype for dtmf in the next major release.

slorello89: Aye, I see that - however since the return type of `getDtmf` is generic, if you don't explicitly tell the compiler what the type is, it can't infer it as a string (hence why this would be a breaking change). I'm not sure if you can do this in java, but you can may be able to keep the original class non-generic and have a new `InputEvent` class that is generic and mark the old class as deprecated altogether?

yallen011: It will still work without specifying the generic type when using getDTMF() for DTMF only the old way. The problem will arise from doing DTMF and/or Speech the new way. I can't do 2 InputEvent files but I can do InputEventv2 but since my concern is about the fact that this is deprecated and will go away eventually anyway so creating a InputEventv2 file only to later remove it and put everything in InputEvent later seems cumbersome. @dragonmantank given the situation, can we go ahead with this being a breaking change? I don't think we can avoid having a breaking change here. No matter what route we choose. All roads lead to a breaking change...whether it be now or later. Once we deprecate, it will be another breaking change so we can do a breaking change to allow support for both for a time then break it again when we remove or we can just make it so we only support new DTMF and only have 1 breaking change for this feature.

dragonmantank: ### tl;dr - If we can't do this without it being a breaking change, remove the old way and cut a major release.

Ultimately any solution we provide we want to give customers the chance to upgrade as soon as possible. We deprecated old functionality and load the new functionality along-side. Existing customers can continue to use the old syntax until they get a chance to migrate. When a user can use new functionality without upgrading major versions, it's a huge win. It's more work on our side to do that, but ultimately we're here to make the developer's lives easier. Upgrading to a new major version isn't always something a developer can just do.

The decision is up to you. If we simply cannot provide a solution that enables the old syntax as well as the new syntax, we'll have to ship it as a breaking change and a new major version. If we're doing the breaking change, just rip out the old DTMF-only Input and replace it with the new DTMF+ASR syntax. 

> All roads lead to a breaking change

You're correct. The question _we_ have to ask ourselves is where do we make that fork in the road. Do we add additional lanes and signage to help developers know that the fork is coming up and give them a chance to change lanes, or do we just plop down an off-ramp. 

> Once we deprecate, it will be another breaking change

Not quite. Deprecating a feature is not removing functionality, so that's a minor change. We're alerting the user that this code path is going away. By providing the new way alongside the deprecated way, we give the user a chance to upgrade. _That might not always be possible_, but that doesn't mean we shouldn't try. 

Deprecated features can, and should, be removed at the next major release. We are then free to remove deprecated features at any major version release, at our discretion. So something deprecated in 5.x can be removed in 6.0.0. Anything deprecated in 6.x can be removed in 7.0.0. That major release can come six months later, or six days later, it doesn't matter. We get the ability to remove functionality only at a major release, so that's when we flush out deprecated features.

yallen011: After talking further with @dragonmantank I have decided to gut out the old DTMF stuff and make it breaking change. This will result in a v6 release. I will submit another PR that will have all the old DTMF stuff removed.

yallen011:
@dragonmantank @slorello89 updated sdk and code snippets to include the dtmf portion of the asr changes.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [6.0.0]
```
slorello89, line 7: Just for clarity - are we going to have a 5.7.0 that has both the new Input stuff and the outmoded DTMF stuff, and then a 6.0.0 that removes it all?

src/main/java/com/vonage/client/incoming/InputEvent.java
```diff
@@ -23,36 +23,79 @@
...
+    /**
+     * @return The unique identifier for this conversation
+     */
     @JsonProperty("conversation_uuid")
     public String getConversationUuid() {
         return conversationUuid;
     }
 
+    /**
+     *Deprecated. Use {@link DtmfResult#isTimedOut()}
```
slorello89, line 54: Just want to make sure this doesn't work it's way in to 6.0.0.

src/main/java/com/vonage/client/voice/ncco/InputAction.java
```diff
@@ -17,51 +17,48 @@
...
+    @JsonProperty(required = true)
+    private Collection<String> type;
+    private DtmfSettings dtmf;
     private Collection<String> eventUrl;
+    private SpeechSettings speech;
     private EventMethod eventMethod;
 
     /**
-     * @deprecated Use {@link Builder}
+     * @param builder  builder to create InputAction object
```
slorello89, line 41: Just want to make sure this doesn't work it's way in to 6.0.0.

slorello89 (approved):
Only comments are flagging the things that I expect to be removed before 6.0.0 is released. Otherwise, I've tested the snippets and they look good.

src/main/java/com/vonage/client/incoming/InputEvent.java
yallen011: good catch. Removed

src/main/java/com/vonage/client/voice/ncco/InputAction.java
yallen011: Not sure what is wrong with this one? Because this is for 6.0.0 release, we are making the InputAction constructor private and the builder will now be used exclusively instead of allowing for the constructor to be used directly.

PR #355, opened 2020-11-10
changed create application request auth

yallen011:
Changed create application request auth to use basic auth based on changes announced by the Product API team

> As part of releasing an internal security update for the Applications Service, we noticed that v2 of the public-facing Application API was accepting API key/secret as request params for authentication. This has been treated as a bug.
Request parameter auth was never intended to be supported and was not documented publicly or internally for v2 of the API. We recommend customers use header auth (basic auth) for this API instead. https://developer.nexmo.com/concepts/guides/authentication#header-based-api-key-and-secret-authentication

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [5.6.0]
+### Added
+- NotifyEvent structure for Notify Actions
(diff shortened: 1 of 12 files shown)


slorello89 (changes requested):
I think this will need to be done for the other Application Methods as well (Get/list/Delete/Update)

yallen011:
@slorello89 good call. I only checked the call referenced in the slack thread. Let me check the others

dragonmantank (approved):
Was able to run all the code snippets. LGTM :+1:

Issue #375, opened 2022-03-14
Support for retrieve outbound pricing for all countries API


kkazak-wr:
Hello,

Is there any plans to add support to retrieve outbound pricing for all countries API?

GET https://rest.nexmo.com/account/get-full-pricing/outbound/:type

see second API here https://developer.nexmo.com/api/pricing?theme=dark

SMadani:
This has been implemented in #381 and will be available in the next release.

PR #381, opened 2022-05-16
Add get-full-pricing endpoint

SMadani:
This PR addresses #375


src/main/java/com/vonage/client/account/PricingRequest.java
```diff
@@ -17,12 +17,18 @@
 
 public class PricingRequest {
     private String countryCode;
+    private ServiceType serviceType;
 
-    public PricingRequest(String countryCode) {
+    public PricingRequest(String countryCode, ServiceType serviceType) {
```
matt-lethargic, line 22: This is a breaking change, is it going to be mitigated somehow?

matt-lethargic (commented):
Looks good, just question about mitigating the breaking change on `PricingRequest`

src/main/java/com/vonage/client/account/PricingRequest.java
SMadani: It's only intended to be used internally so it's not a major issue. It will probably be part of a major version release anyway.

PR #382, opened 2022-05-17
Renamed 6.x to main

SMadani:
This PR changes references to the 6.x branch to `main` so that in the future this won't be needed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build Java CI
 on:
   push:
     branches:
-      - 6.x
+      - main
   pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build Java CI
 on:
   push:
     branches:
-      - 6.x
+      - main
```
slorello89, line 5: @dragonmantank 😆

SMadani: I'm guessing there's a reason for this? 😅

dragonmantank: I don't make the naming decisions on branches anymore :P

PR #383, opened 2022-05-18
Remove redundant "Endpoint" wrapping

SMadani:
At present, the structure of the SDK is a top-level `VonageClient`, which is composed of multiple sub-clients for each API. Each API's client is then composed of one or more _endpoints_, and each endpoint is composed of one or more _methods_. This latter decomposition is redundant and somewhat misleading. For example, `com.vonage.client.application.ApplicationClient` has a single field: `com.vonage.client.application.ApplicationEndpoint`. This endpoint is just a wrapper around the actual endpoints in the Application API, which are called "Method"s. This additional layer of wrapping is completely unnecessary, adding no value and creates unnecessary duplication and coupling. Since these are package-private classes, there is no risk to breaking compatibility, as users rely on the public API of the "clients", not "endpoints" or "methods". For consistency with all packages in the repo and to avoid confusion, the "Method" classes have been renamed to "Endpoint", and the existing endpoint wrappers have been deleted. This makes the structure flatter and easier to understand, test and maintain.


maxkahan (approved):
Can't fault your logic here - the extra step is unnecessary

PR #384, opened 2022-05-18
Remove redundant enum parsing code

SMadani:
Most enums in the SDK create a map from the name to the enum itself to make it possible to get an instance of the enum from a String. This functionality is already provided by the `valueOf` built-in method on all enum types, so the Map is unnecessary. The PR also removes unused loggers in the endpoint classes.


maxkahan (approved):
lgtm (in my limited experience)

PR #389, opened 2022-06-28
Update Insight to be consistent with current API

SMadani:
This PR addresses the inconsistencies between the [Number Insight API spec](https://developer.vonage.com/api/number-insight) and the Java SDK's implementation of it. Further quality enhancements, such as testing, validation and refactoring are also included in this.


src/main/java/com/vonage/client/insight/Reachability.java
```diff
@@ -0,0 +1,44 @@
+/*
+ *   Copyright 2020 Vonage
```
dragonmantank, line 2: Noticed on a few other new files, this should be 2022

src/main/java/com/vonage/client/insight/StandardInsightResponse.java
```diff
@@ -41,33 +49,97 @@ public static StandardInsightResponse fromJson(String json) {
...
         return refundPrice;
     }
+
+    /**
+     * @return Full name of the person or business who owns the phone number, or "unknown" if this
+     * information is not available. This parameter is only present if cnam had a value of
+     * <code>true</code> in the request.
+     */
+    @JsonProperty("caller_name")
+    public String getCallerName() {
```
dragonmantank, line 116: This might be something to be brought up to the VAPI team, but why are we exposing this information twice? Seems odd that we have both a `CallerIdentity` and this in the responses (I know the SDK is just reflecting the actual response, but should these direct exposures be deprecated in the API?) Might be useful to get that answered before we add new code that might turn around and be deprecated.

dragonmantank (changes requested):
Couple of changes/questions

src/main/java/com/vonage/client/insight/Reachability.java
SMadani: Everything in the repo except Messages API is 2020 so I'll open a separate PR to update these

dragonmantank: +1

src/main/java/com/vonage/client/insight/StandardInsightResponse.java
SMadani: I agree, this puzzled me too. Same with the [`handset_status`](https://vonage.slack.com/archives/CD2JDM09M/p1656670938668069), which has undocumented values yet could very well be be a Boolean.

src/main/java/com/vonage/client/insight/AdvancedInsightRequest.java
```diff
@@ -16,26 +16,57 @@
...
+
+    public Boolean getRealTimeData() {
+        return realTimeData;
+    }
+
     public Boolean getCnam() {
         return cnam;
     }
 
+    @Deprecated
```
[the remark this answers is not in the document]
dragonmantank: If we can't do a final 6.x release then no big deal, we can leave the deprecation here. Just would be nice to excise it now properly, but still alert the users.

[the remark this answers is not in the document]
SMadani: Well, we didn't deprecate the Search API in the SDK and now it's completely gone. In an ideal world deprecations would happen before removals I agree. Though perhaps in this case it is better to leave it as deprecated since the API itself marks it as deprecated, not removed.

dragonmantank (approved):
LGTM

PR #390, opened 2022-07-04
Move Channel and MessageType to Request constructor

SMadani:
This PR moves the Channel and Message type in the Messages implementation to be constructor parameters in MessageRequests, which is more appropriate and simpler than having builders override an abstract method to set it, since it's immutable by design.


maxkahan (approved):
I see your logic

PR #391, opened 2022-07-05
Be more permissive with SMS and MMS senders

SMadani:
This PR allows any string to be used for the sender field by default in `MessageRequest`, since the [API spec ](https://developer.vonage.com/api/messages-olympus) does not strictly prohibit non-E164 numbers to be used as the `from` field in SMS and MMS requests.


maxkahan (approved):
LGTM - makes sense to make it broader so customers can use a pattern they're more familiar with

PR #394, opened 2022-07-10
Ensure User-Agent is set in all request headers and UTF-8 encoding is used

SMadani:
This PR is a "just in case" fix for the User-Agent potentially not being set in requests, including a test for it. It also ensures the UTF-8 charset is used, since it appears to be an unresolved bug in `org.apache.http.client.methods.RequestBuilder` constructors.


maxkahan (approved):
Your logic makes sense to me

PR #395, opened 2022-07-14
Deprecate Redact client

SMadani:
The [Redact API](https://developer.vonage.com/api/redact) was never GA's (or even enter Beta stage). This PR deprecates it for clarity, with the aim of removing it in a future release.


maxkahan (approved):
Have read and all LGTM!

PR #397, opened 2022-07-18
Remove VoiceName

SMadani:
This PR removes the deprecated `voiceName` field to discourage its use, since it has been deprecated for quite some time now. See [the documentation](https://developer.vonage.com/voice/voice-api/guides/text-to-speech#legacy-voice-names) for context.


devwithzachary (approved):
All looks good!

PR #399, opened 2022-07-20
Update SMS status codes

SMadani:
The current `com.vonage.client.sms.MessageStatus` enum is missing error codes and has some that are no longer documented / in use. This PR adds the missing ones, deprecates the unused and also adds Javadocs.


dragonmantank (approved):
LGTM

PR #403, opened 2022-08-09
Use Vonage JWT library

SMadani:
This PR updates the dependency on [nexmo-jwt-jdk](https://github.com/Nexmo/nexmo-jwt-jdk) to use [vonage-jwt-jdk](https://github.com/Vonage/vonage-jwt-jdk).


pardel (approved):
LGTM

PR #404, opened 2022-08-10
Fix WhatsappTemplate

SMadani:
This PR fixes some inaccuracies / inconsistencies in `WhatsappTemplateRequest`. Namely, it changes `parameters` to be a `List<String>` instead of `List<?>`, makes `policy` optional and adds enumerations for supported locales. It also restricts the sender field to numbers.

### DIFF
--- a/src/main/java/com/vonage/client/messages/messenger/MessengerAudioRequest.java
+++ b/src/main/java/com/vonage/client/messages/messenger/MessengerAudioRequest.java
@@ -22,7 +22,7 @@ import com.vonage.client.messages.MessageType;
 
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 public final class MessengerAudioRequest extends MessengerRequest {
-	MessagePayload audio;
+	final MessagePayload audio;
 
 	MessengerAudioRequest(Builder builder) {
 		super(builder, MessageType.AUDIO);
--- a/src/test/java/com/vonage/client/messages/whatsapp/WhatsappTemplateRequestTest.java
+++ b/src/test/java/com/vonage/client/messages/whatsapp/WhatsappTemplateRequestTest.java
@@ -26,7 +26,7 @@ public class WhatsappTemplateRequestTest {
 	public void testSerializeMandatoryFields() {
 		String json = WhatsappTemplateRequest.builder()
 				.from("Acme Corp").to("447900000001")
-				.name("verify").locale("en-GB")
+				.name("verify").locale("en-US")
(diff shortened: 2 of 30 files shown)


src/test/java/com/vonage/client/messages/whatsapp/WhatsappTemplateRequestTest.java
```diff
@@ -26,90 +26,69 @@ public class WhatsappTemplateRequestTest {
...
+		assertTrue(json.contains("\"whatsapp\":{\"locale\":\"en\"}"));
 	}
 
 	@Test
 	public void testSerializeAllFields() {
 		String json = WhatsappTemplateRequest.builder()
-				.from("Acme Corp").to("447900000001")
-				.name("verify").locale("en-GB").policy(Policy.DETERMINISTIC)
-				.parameters(Arrays.asList(Collections.singletonMap("k1", "v1"), "blah"))
+				.from("Acme Corp").to("447900000001").name("verify")
```
superchilled, line 42: For WhatsApp, `from` would be a number. Though this is just a test, so I guess not a huge issue using a name here.

src/main/java/com/vonage/client/messages/messenger/MessengerAudioRequest.java
```diff
@@ -22,7 +22,7 @@
 
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 public final class MessengerAudioRequest extends MessengerRequest {
-	MessagePayload audio;
+	final MessagePayload audio;
```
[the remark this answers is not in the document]
SMadani: It means the value can't be changed, so must be set in the constructor.

src/test/java/com/vonage/client/messages/whatsapp/WhatsappTemplateRequestTest.java
SMadani: So `from` is always a number? The spec didn't make that clear, I'll have to add some validation to it then.

superchilled: Before you add validation, let me just double-check that it can never be text/name.

superchilled: @SMadani got confirmation from Gulam that `from` is always a number for WhatsApp.

superchilled (approved):

PR #406, opened 2022-08-22
Fix parsing MessageResponseException with no body

SMadani:
This PR fixes parsing error responses from the Messages v1 API ([as reported on the Community Slack](https://vonage-community.slack.com/archives/C9H152ATW/p1660931679229029)) where the response doesn't contain JSON body, leading to `VonageUnexpectedException` being thrown as the response is attempted to be parsed into the `MessageResponseException` object.


dragonmantank (approved):
LGTM

Issue #405, opened 2022-08-17
Wrong Content-Type for Verify requests


pablosantiago:
**Describe the bug**
When using the Java SDK to make requests to the Verify API, the calls fails due to wrong `Content-Type` header.
`VerifyEndpoint` creates a HTTP request with `Content-Type` set as `application/json` but adds all parameters as URL parameters. No JSON request body is created.
This results in the following response:

```
{
    "status": "999",
    "error_text": "Could not parse request :: Unparseable JSON"
}
```

**To Reproduce**
```java
final VonageClient client = VonageClient.builder()
        .apiKey("XXXXX")
        .apiSecret("XXXXX")
        .build();

final VerifyClient verifyClient = client.getVerifyClient();
final VerifyRequest request = VerifyRequest.builder("447700900000", "AcmeInc").build();

final VerifyResponse response = verifyClient.verify(request);
System.out.println("response = " + response.getErrorText());
```

**Expected behavior**
A successful response.

**Request/Response**
```
curl --location --request POST 'https://api.nexmo.com/verify/json?api_key=XXXXXX&api_secret=XXXXXXX&number=447700900000&brand=AcmeInc' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json'
```
```
{"status":"999","error_text":"Could not parse request :: Unparseable JSON"}
```

**Vonage SDK Version**
 - Version: 7.0.0

**Java SDK Version**
- Version: 11

**Proxy Server Used**
 * [ ] Yes
 * [ X ] No

pablosantiago:
Removing the `Content-Type` header (still a POST request with `Accept=application/json`), results in:
```<html>

<head>
	<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1" />
	<title>Error 400 </title>
</head>

<body>
	<h2>HTTP ERROR: 400</h2>
	<p>Problem accessing /verify/json. Reason:
		<pre>    Bad Request</pre>
	</p>
	<hr /><i><small>Powered by Jetty://</small></i>
</body>

</html>
```

However, changing it to a GET request works.
According to the [documentation](https://developer.vonage.com/api/verify#requests), both GET and POST should work.

pablosantiago:
It works with version 6.4.2 but not with > 6.5.0.

xejarque:
Exactly the same issue upgrading to version 7.0.0

SMadani:
Fixed in 7.1.0 (available now).

PR #407, opened 2022-08-22
Update Verify

SMadani:
Seems like the `require_type` parameter is no longer used as it's missing from [the spec](https://developer.vonage.com/api/verify#requests). This PR also fixes the incorrect `Content-Type` header for Verify requests (see #405).

### DIFF
--- a/src/main/java/com/vonage/client/verify/BaseRequest.java
+++ b/src/main/java/com/vonage/client/verify/BaseRequest.java
@@ -27,7 +27,7 @@ import java.util.Locale;
  * Base request class for {@link VerifyRequest} and {@link Psd2Request}
  * @since 5.5.0
  */
-public class BaseRequest {
+public abstract class BaseRequest {
     private final String number;
     private Integer length;
     private Locale locale;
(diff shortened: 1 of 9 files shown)


src/main/java/com/vonage/client/verify/BaseRequest.java
```diff
@@ -27,7 +27,7 @@
  * Base request class for {@link VerifyRequest} and {@link Psd2Request}
  * @since 5.5.0
  */
-public class BaseRequest {
+public abstract class BaseRequest {
```
SecondeJK, line 30: wow, how was this not the case before

src/main/java/com/vonage/client/verify/VerifyEndpoint.java
```diff
@@ -41,43 +40,9 @@ protected Class<?>[] getAcceptableAuthMethods() {
     public RequestBuilder makeRequest(VerifyRequest request) throws UnsupportedEncodingException {
         RequestBuilder result = RequestBuilder
                 .post(httpWrapper.getHttpConfig().getApiBaseUri() + PATH)
-                .setHeader("Content-Type", "application/json")
```
SecondeJK, line 44: This is all being killed to simplify the request object, right?

SMadani: The header was wrong (that's the line you highlighted)

SecondeJK (approved):

PR #429, opened 2022-12-14
Add client token generation for Video API

SMadani:
This PR adds the missing token generation capabilities to the SDK, which is required for the Video integration. Users call the `VideoClient#generateToken(String, TokenOptions)` method to get a new signed JWT for a given session with the desired options.
Implementation-wise, this piggybacks on the `JWTAuthMethod`, which uses the [`com.vonage:jwt` library](https://github.com/Vonage/vonage-jwt-jdk).


dragonmantank (approved):
LGTM

PR #436, opened 2023-02-09
Add Video API SIP endpoints

SMadani:
Implements [SIP functionality for the video API](https://nexmo-api-spec-pr-586.herokuapp.com/video#sip).

### DIFF
--- a/src/main/java/com/vonage/client/video/SendDtmfToSessionEndpoint.java
+++ b/src/main/java/com/vonage/client/video/SendDtmfToSessionEndpoint.java
@@ -1,0 +1,58 @@
+/*
+ *   Copyright 2023 Vonage
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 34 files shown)


src/main/java/com/vonage/client/video/SendDtmfToSessionEndpoint.java
```diff
@@ -0,0 +1,58 @@
...
+	public RequestBuilder makeRequest(SendDtmfRequest request) {
+		String path = String.format(PATH, getApplicationIdOrApiKey(), request.sessionId);
+		String uri = httpWrapper.getHttpConfig().getVideoBaseUri() + path;
+		return RequestBuilder.post(uri)
+				.setHeader("Content-Type", "application/json")
+				.setEntity(new StringEntity(request.toJson(), ContentType.APPLICATION_JSON));
+	}
+
+	@Override
+	public Void parseResponse(HttpResponse response) throws IOException {
```
Tr00d, line 52: I saw this exact method a lot. Maybe you could try to reduce the duplication by centralizing the logic somewhere.
There's nothing specific for SendDtmfToSession in there

Tr00d (approved):

src/main/java/com/vonage/client/video/SendDtmfToSessionEndpoint.java
SMadani: This is the reduced logic. It's the same everywhere in the SDK. The specific part is the path, content-type and response return type as well as the request.

PR #439, opened 2023-03-08
Update copyright headers

SMadani:
Updates the copyright headers and fixes some minor Javadoc formatting issues.


SecondeJK:
One thing that is forthcoming for the PHP SDK is removing the copyright, as it's covered by the license and the composer file - not sure if Gradle/Maven have the same ability to hold copyright though

SecondeJK (approved):

PR #449, opened 2023-04-14
Add duration and file_size to Viber Video message

SMadani:
Duration and file size are required fields according to [the API spec](https://developer.vonage.com/en/api/messages-olympus).

### DIFF
--- a/src/main/java/com/vonage/client/messages/viber/ViberService.java
+++ b/src/main/java/com/vonage/client/messages/viber/ViberService.java
@@ -21,25 +21,33 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 @JsonInclude(value = JsonInclude.Include.NON_NULL)
 public final class ViberService {
 	private final Category category;
-	private final Integer ttl;
+	private final Integer ttl, duration, fileSize;
 	private final String type;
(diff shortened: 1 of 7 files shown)


src/main/java/com/vonage/client/messages/viber/ViberService.java
```diff
@@ -21,25 +21,33 @@
...
-		if (category == null && ttl == null && type == null && action == null) {
+	static ViberService construct(Category category, Integer ttl, String type, Action action,  Integer duration, Integer fileSize) {
+		if (category == null && ttl == null && type == null && action == null && duration == null && fileSize == null) {
 			return null;
 		}
 		if (ttl != null && (ttl < 30 || ttl > 259200)) {
 			throw new IllegalArgumentException("Time-to-live (ttl) must be between 30 and 259200 seconds");
 		}
-		return new ViberService(category, ttl, type, action);
+		if (duration != null && (duration < 1 || duration > 600)) {
```
SecondeJK, line 44: not sure about Java patterns in general, but would it be worth having a validation handler here now that the constructor is doing a lot more of it?

SecondeJK (approved):

PR #466, opened 2023-07-14
Reduce boilerplate and decouple endpoints from Apache HttpClient

SMadani:
This PR is an attempt to standardise the implementation of endpoints whilst reducing the repetition and coupling involved. Most of our APIs (especially newer ones) follow a very similar structure: JSON request & response payloads, a predictable exception handling approach and boilerplate implementations for tests. Instead of repeating the same implementation with a few tweaked parameters everywhere, the maintainability of the SDK will be greatly enhanced if there is reduced duplication and decreased coupling from the underlying HTTP client library.

The approach I've taken is to introduce `DynamicEndpoint` along with `Jsonable` and `QueryParamsRequest`. These classes standardise the implementation of endpoints and request / response objects. In particular, Dynamic Endpoints mean that a declarative approach to implementing endpoints can be taken rather than an imperative one. This allows for not only more concise implementation of new APIs and endpoints, but also one which is decoupled from the internals of the HTTP library. Consequently this also makes testing more declarative, hence the introduction of `DynamicEndpointTestSpec`. To test this approach, I have reimplemented Verify v2, Subaccounts, Messages, SMS, SNS and Redact using dynamic endpoints. This demonstrates that the approach is generalisable not only to new APIs but also legacy ones which are less standardised. Further improvements and refactoring will be made in a future PR. This lays the groundwork.


superchilled (approved):
All looks good, as discussed on the call.

PR #477, opened 2023-08-28
Refactor Numbers and Conversion clients to use DynamicEndpoint

SMadani:
This continues the work of moving to a more declarative way of implementing and testing endpoints by decoupling from Apache HttpClient. This PR targets the Numbers and Conversion APIs.


maxkahan (approved):
I love it, I'll take 100

PR #501, opened 2023-12-13
feat: Add Verify2 SMS from field

SMadani:
Adds the `from` field to `com.vonage.client.verify2.SmsWorkflow`.


SecondeJK (changes requested):
Sorry I should have added in the requirements that from is optional - and I can't -see- any optional modifiers here

SMadani:
> Sorry I should have added in the requirements that from is optional - and I can't -see- any optional modifiers here

It is optional - there are two other constructors that don't use `from` and it can be null too.

SecondeJK (approved):

PR #504, opened 2024-01-02
feat: Add new properties to Messages API

SMadani:
Various updates to reflect recent changes in the [Messages v1 spec](https://developer.vonage.com/en/api/messages-olympus). Description of these is in each commit headline.

### DIFF
--- a/src/main/java/com/vonage/client/messages/whatsapp/Context.java
+++ b/src/main/java/com/vonage/client/messages/whatsapp/Context.java
@@ -57,7 +57,10 @@ public final class Context {
 	 * Only applies to Order messages.
 	 *
 	 * @return The referred product details, or {@code null} if not applicable.
+	 *
+	 * @deprecated This will be moved in a future release.
 	 */
+	@Deprecated
 	@JsonProperty("whatsapp_referred_product")
 	public ReferredProduct getReferredProduct() {
 		return referredProduct;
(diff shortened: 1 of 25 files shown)


src/main/java/com/vonage/client/messages/whatsapp/Context.java
```diff
@@ -57,7 +57,10 @@ public UUID getMessageUuid() {
 	 * Only applies to Order messages.
 	 *
 	 * @return The referred product details, or {@code null} if not applicable.
+	 *
+	 * @deprecated This will be moved in a future release.
```
superchilled, line 61: Do you mean 'removed'?

```suggestion
	 * @deprecated This will be removed in a future release.
```

superchilled (commented):
Looks good overall. Just wanted to check a couple of things because I couldn't see them in the changed files:
- For WhatsApp Inbound (all message types) add `context.message_uuid` and `context.message_from`
- For Viber Outbound (type `video`) add `viber_service.duration` and `viber_service.file_size`

Maybe you already updated these properties prior to this PR?

src/main/java/com/vonage/client/messages/whatsapp/Context.java
SMadani: No I mean moved, because it's now part of the `whatsapp` object in the spec.

SMadani:
> Looks good overall. Just wanted to check a couple of things because I couldn't see them in the changed files:
> 
> * For WhatsApp Inbound (all message types) add `context.message_uuid` and `context.message_from`
> * For Viber Outbound (type `video`) add `viber_service.duration` and `viber_service.file_size`
> 
> Maybe you already updated these properties prior to this PR?

Yep, they've been in the spec for a while and hence the SDK :)

superchilled (approved):

PR #507, opened 2024-01-10
feat: Add Number Insight v2

SMadani:
Implements [NI v2 spec](https://developer.vonage.com/en/api/number-insight.v2).

### DIFF
--- a/src/main/java/com/vonage/client/numberinsight2/FraudCheckRequest.java
+++ b/src/main/java/com/vonage/client/numberinsight2/FraudCheckRequest.java
@@ -1,0 +1,66 @@
+/*
+ *   Copyright 2024 Vonage
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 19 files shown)


maxkahan (approved):

src/main/java/com/vonage/client/numberinsight2/FraudCheckRequest.java
```diff
@@ -0,0 +1,66 @@
...
+	private final Set<Insight> insights;
+
+	public FraudCheckRequest(String phoneNumber, Insight insight, Insight... others) {
+		type = "phone";
+		phone = new E164(phoneNumber).toString();
+		insights = EnumSet.of(Objects.requireNonNull(insight, "Insight type is required."), others);
+	}
+
+	/**
+	 * Accepted value is “phone” when a phone number is provided.
```
maxkahan, line 38: The quotes on this line are causing an error on Windows

maxkahan (changes requested):

src/main/java/com/vonage/client/numberinsight2/FraudCheckRequest.java
SMadani: Good catch! Thanks

maxkahan (approved):

PR #518, opened 2024-03-05
feat: Add Conversations v1

SMadani:
This PR implements the [Conversation API](https://developer.vonage.com/en/api/conversation), along with some tidying up.


SMadani:
jwtk/jjwt#932  raised.

PR #547, opened 2024-10-20
Fix Whatsapp API `latitude` and `longitude` field names

jeremychan:
According to the API https://developer.vonage.com/en/messages/code-snippets/whatsapp/send-location for sending location the field names should be "latitude" and "longitude". Currently "lat" and "long" is causing the message to not deliver properly


SMadani:
Thank you for reporting this and an attempt to fix. I can't merge this as-is since the tests need updating, and `lat` / `long` is used for deserialisation of inbound messages (as per [the spec](https://developer.vonage.com/en/api/messages#inbound-message)), so will need to use `@JsonAlias` instead.

SMadani:
Fixed in 2bf35e0f8c9435488a4fa50b384a882bf6ae721e.

jeremychan:
thank you

PR #615, opened 2026-03-05
RCS fixes

dragonmantank:
* Added some missing RCS config data
* Made TTL public so it can be set

## Contribution Checklist
* [x] Unit tests!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 # [9.10.0]
-<<<<<<< HEAD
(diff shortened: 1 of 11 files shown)


superchilled (changes requested):
Looks good overall. Just one suggestion re the `ttl` bounds. 

However, looking at the tests you added I did notice another issue. The `trustedRecipient` [param](https://github.com/Vonage/vonage-java-sdk/blob/263fb2b6de09e4ad18687caef57a998a826db392/src/main/java/com/vonage/client/messages/rcs/Rcs.java#L51) should be a *top-level* param, not part of the `rcs` object (see the [spec](https://developer.vonage.com/en/api/messages#SendMessage)).

Would probably make sense to fix that as part of this same PR.

PR #627, opened 2026-08-04
feat: WhatsApp BSUID support in Messages API (DEVX-11482)

dragonmantank:
## Summary

Adds support for WhatsApp **business-scoped user IDs (BSUIDs)** to the Messages API. When a WhatsApp user adopts a username, Meta may identify them by a BSUID instead of (or in addition to) a phone number. This PR lets the SDK send to BSUID recipients and parse the new BSUID fields on inbound messages and status callbacks.

Reference: [Handling WhatsApp Usernames and BSUIDs](https://developer.vonage.com/messages/guides/handling-whatsapp-usernames-and-bsuids)

## Changes

**Outbound (send)**
- WhatsApp `to` now accepts a BSUID or parent BSUID as well as an E.164 phone number.
- New `Bsuid` validation helper — format `{ISO alpha-2}.{≤128 alphanumeric}` (e.g. `US.13491208655302741918`) and parent BSUID `{CC}.ENT.{id}` (e.g. `US.ENT.11815799212886844830`).
- `WhatsappRequest` recipient validation relaxed via `sanitizeRecipient()` (accepts E.164 **or** BSUID); `from` remains E.164.

**Inbound webhook**
- New `whatsapp.sender` object exposed via `InboundMessage.getWhatsappSender()`.
- `Profile.getUsername()` added.

**Status callback**
- New `whatsapp.recipient` object exposed via `MessageStatus.getWhatsappRecipient()`.
- WhatsApp `profile` exposed via `MessageStatus.getWhatsappProfile()` (includes username).

**Shared model**
- New `WhatsappUser` (`user_id` = BSUID, `parent_user_id` = parent BSUID, `wa_id` = phone number) used by both `whatsapp.sender` and `whatsapp.recipient`.

## Tests
- New `BsuidTest` (valid/parent/max-length/malformed/null cases).
- BSUID send + parent BSUID + sender-cannot-be-BSUID + invalid-recipient cases in `WhatsappTextRequestTest`.
- `whatsapp.recipient` + `profile.username` parsing in `MessageStatusTest`.
- `whatsapp.sender` + `profile.username` parsing in `InboundMessageTest`.
- All 332 messages-package tests pass; a full `mvn test` run passed (1465 tests, 0 failures).

## Notes
- `@since` tags use `9.13.0` (next version after the released 9.12.0). `pom.xml` version bump not included — assumed handled at release time via `bumpversion.sh`.
- BSUIDs cannot be used for one-tap/zero-tap/copy-code auth templates (server-enforced, error 131062); documented in the builder Javadoc, not enforced client-side.


superchilled (commented):
Looks good overall. My only question is whether we maybe want to be more permissive wrt the `to` field, e.g. allow any string input regardless of format. The current logic of `sanitizeRecipient` checks if the input is either in E.164 format or matches the defined regex for BSUID format.

This is fine given the current known parameters of what can be specified as the recipient for a WhatsApp message, but:

1. We know from previous experience that Meta has a habit of regularly changing their API and definitions, so the current BSUID format may be subject to change in the future
2. Other valid formats for the `to` field will require support in the future. For example, we already know that [WhatsApp group messaging](https://developers.facebook.com/documentation/business-messaging/whatsapp/groups) is on the Messages API product roadmap, and the group ID again has a [different format](https://developers.facebook.com/documentation/business-messaging/whatsapp/groups/reference#request-parameters-1)

dragonmantank:
Updated per review feedback (commit `ec188406`):

Dropped the bespoke `Bsuid` format validation on `to`. Rationale: a strict client-side regex risks rejecting recipients that Meta actually accepts whenever they loosen their format, which would force an urgent SDK release. 

New behaviour for WhatsApp `to`:
- If it parses as an E.164 phone number, it's normalised (existing behaviour — strips spaces/dashes/leading `+`).
- Any other non-empty value (e.g. a BSUID, or a future provider format) is passed through **unchanged**.
- `null`/empty are still rejected; `from` is still strictly E.164.

The API validates the recipient and returns error `131062` for genuinely unsupported values, so we lose nothing by not duplicating that check client-side. Removed `Bsuid.java`/`BsuidTest.java`; added passthrough + normalisation tests. All 328 messages-package tests pass.

superchilled (approved):

